PR #88, opened 2014-11-25
Add cljr-add-missing-libspec

expez:
`cljr-add-missing-libspec` will try to resolve the symbol at point and
require or import the missing var.  If there are more than one var match
the user is prompted. If there's only one result it is added without
user intervention.  If the symbol at point is of the form
`edn/read-string` the resulting libspec will be aliased to `edn`.

This depends on the middleware to do its thing.

action shot

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,8 @@ This is it so far:
 
  - `ad`: add declaration for current top-level form
  - `ai`: add import to namespace declaration, then jump back
+ - `ap`: add a dependency to your project **depends on refactor-nrepl 0.1.0 and above**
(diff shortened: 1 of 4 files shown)


arichiardi:
By the way what is the package that I need to install in order to ```cljr-aml``` instead of ```cljr-add-missing-spec``` ?

magnars:
smex

arichiardi:
:bow:

expez:
I used `flx-ido` when recording that gif.

arichiardi:
Great, at the moment I am using ```helm``` for commands...I am checking if it has that nice feature...

PR #103, opened 2014-12-28
Add cljr-extract-function

### LINKED ISSUE #11: extract fn / inline fn / rename
- extract defn or defn-
- inline function
- rename within project

All the hard ones. Even thought they're relatively harder, I'd like to talk more about how we could implement them here.

expez:
Extract the form at point, or the nearest enclosing form, into a toplevel defn.

With a prefix the newly created defn will be public.

This depends on the [find-unbound](https://github.com/clojure-emacs/refactor-nrepl/pull/15) op in the middleware.

Fixes #11

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,7 @@ This is it so far:
  - `am`: add a missing libspec
  - `ap`: add a dependency to your project
  - `cn`: Perform various cleanups on the ns form
+ - `ef`: Extract function
  - `fu`: Find usages
  - `hd`: Hotload dependency
  - `rd`: Remove (debug) function invocations
@@ -548,6 +549,13 @@ Note that we have to build an AST in order to do this, so if the current file is
 
 If point is on a dependency vectors in, e.g. in `project.clj`, it will get hotloaded into the repl.
 
+
+## Extract function
+
+Extract the form at point, or the nearest enclosing form, into a toplevel defn.
+
+With a prefix the newly created defn will be public.
+
 ## Miscellaneous
 
 With clj-refactor enabled, any keybindings for `paredit-raise-sexp` is
@@ -562,6 +570,7 @@ You might also like
 
 ## Changelog
 
+- Add `cljr-extract-function` [Lars Andersen](https://github.com/expez)
 - Add `cljr-hotload-dependency` [Lars Andersen](https://github.com/expez)
 - Hotloading of artifacts added with `cljr-add-project-dependency` [Lars Andersen](https://github.com/expez)
 - Add `cljr-remove-let` [Lars Andersen](https://github.com/expez)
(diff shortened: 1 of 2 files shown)


magnars:
It's great that you've tackled this feature :-)

One issue I've found: It includes namespace local functions as arguments. 

In my case, it looked like this:

```clj
(defn get-badge-jump-value [badge]
  (case (:type badge)
    :bronze 5
    :silver 10
    :gold 20))

(defn find-forward-time-events [[type & info]]
  (case type
    :added-badge (let [b (badge (first info))]
                   [:added-time-event {:event :new-badge
                                       :type :gain
                                       :magnitude (get-badge-jump-value b)
                                       :details {:badge-id (:id b)
                                                 :quality (:type b)}}])
    nil))
````

Doing `extract-function` on the let, I get:

```clj
(defn- added-badge-event
  [badge info get-badge-jump-value]
  (let [b (badge (first info))]
                   [:added-time-event {:event :new-badge
                                       :type :gain
                                       :magnitude (get-badge-jump-value b)
                                       :details {:badge-id (:id b)
                                                 :quality (:type b)}}]))

(defn find-forward-time-events [[type & info]]
  (case type
    :added-badge (added-badge-event badge info get-badge-jump-value)
    nil))
```

Notice how `get-badge-jump-value` is included in the parameter list, even tho it is locally bound in the namespace.

magnars:
I am also unable to extract just the vector `[:added-time-event ...]`, it jumps up and includes the let either way.

expez:
Thanks for trying this out.  I should probably only hack on `clj-refactor` when I'm working on another clj project simultaneously so I can dogfood new features :p 

The way this is implemented now is that the form at (or around) point is analyzed and any unbound vars become function parameters.  What we probably want to do is extract the form out into a new function, analyze the entire file up to and including the new function, and let anything unbound become the function parameters.

The problem with vector is here:

```elisp
  (unless (looking-at "(")
    (paredit-backward-up))
```

What we need is a `cljr--looking-at-sexp`.  Maybe something like that already exists in paredit?

expez:
I think these [two](https://github.com/clojure-emacs/clj-refactor.el/commit/5e3774420d2515b17adc91bdac0cf92c83913042) [commits](https://github.com/clojure-emacs/clj-refactor.el/commit/e043cb0566d1f3c5e64b27b587a14e74bc18191b) solve the problems above.

magnars:
@expez The vector extraction now works, but I'm still getting functions that are available in the namespace passed in as arguments.

expez:
Think I need to talk to Bronsa on IRC to get this to work properly.  Working with 'bad code' is a bit tricky and `tools.analyzer` has quite a few extension points.

expez:
[This](https://github.com/clojure-emacs/clj-refactor.el/commit/80df6cc8257afa213eee166ed764a6f297fc5a1b) should take care of it.  The commit message contains a (minor) caveat.

I've pushed a new version to clojars with the matching refactor-nrepl changes.

Issue #58, opened 2014-05-20
customizable magic requires


kdavy:
I love the magic requires feature, but I would love it even more if I could easily customize the list of magic namespaces.

AeroNotix:
Closable?

magnars:
Aye!

PR #112, opened 2015-01-31
Allow for custom magic requires

AeroNotix:


AeroNotix:
Travis failure looks unrelated.

magnars:
It seems wasteful to calculate optimal regexen every time the function is called. How about a `cljr-set-magic-requires` fn instead?

clj-refactor.el
```diff
@@ -1511,7 +1511,7 @@ front of function literals and sets."
   (interactive)
   (insert "/")
   (when (and cljr-magic-requires
-             (looking-back cljr--magic-requires-re 6))
+             (looking-back (cljr--magic-requires-re) 6))
```
expez: This hardcoded 6 seems a bit weird.  Shoudn't that be something like `point-at-bol` instead of the buffer position 6?

expez:
Thanks for taking a look at this, be nice to finally close #58

AeroNotix:
@magnars I was thinking this however:

* Adding an API to such a trivial operation (adding to a list) seems a bit unusual.
  * Users would be more familiar with the `add-to-list` and `push` functions
* Calculating it on the fly is not *that* expensive, so with keeping a nice API and not
being slow. I thought it was fine.


Thoughts?

clj-refactor.el
AeroNotix: I don't know what the 6 is for. Care to enlighten me?

expez: It's used to limit the search to make it faster.  When this code was written I think the author meant it as "look back no more than 6 characters", but what it really means is "when looking back, look at the region between buffer position 6 and the position of point"

magnars: Yeah, I was confused, since I have this function in expand-region:

```el
(defun er/looking-back-max (regexp count)
  "Version of `looking-back' that only check COUNT chars back."
  (looking-back regexp (max 1 (- (point) count))))
```

which gives the API I thought I was using. :P

magnars:
@AeroNotix I think your arguments are compelling. I change my mind. :)

clj-refactor.el
AeroNotix: Do you want me to deal with this in this PR?

magnars: I'll clean it up after merge.

AeroNotix:
@magnars awesome, anything left to do then?

magnars:
Just checking the travis build. Seems like it was just a connection timeout. :+1:

AeroNotix:
Thanks a lot @magnars !

magnars:
And thank you. :)

PR #114, opened 2015-03-05
Fix add-missing-libspec with records and deftypes

expez:
This also fixes an assumption that anything with capitalized names were
classes, which bit me when using prismatic/schema.

The middleware had to be changed to return more information.  We have to
know whether what we're importing is a missing ns, a missing
deftype/defrecord or a missing class.  It now returns an alist of the
following form ((candidate1 . type1)  (candidate2 . type2)..) where
:type is :record, :type, :class or :ns.  When we're importing interfaces
a class is generated and the import mechanism the same so no effort is
made to distinguish between tehse cases.

The case of missing types and records is most complex because we have to
both require the ns, in order to trigger compilation, and then import
the resulting class.

### DIFF
--- a/features/step-definitions/clj-refactor-steps.el
+++ b/features/step-definitions/clj-refactor-steps.el
@@ -66,20 +66,24 @@
     (run-rename-symbol "example.two" '(6  7  1  10  "star*"  "tmp/src/example/two.clj"  "" 8  8  17  27  "star*"  "tmp/src/example/one.clj"  "") "asterisk*")))
 
 (Given "^I call the add-missing-libspec callback directly with mock data to import"
-  (lambda ()
-    (cljr--add-missing-libspec "Date" "java.util.Date" "import")))
+       (lambda ()
+         (cljr--add-missing-libspec "Date" '((java.util.Date :class)))))
 
 (Given "^I call the add-missing-libspec callback directly with mock data to refer split"
-  (lambda ()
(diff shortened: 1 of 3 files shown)


expez:
This functionality will break unless you update `clj-refactor` while grabbing the snapshot of `refactor-nrepl` I'm about to publish

features/step-definitions/clj-refactor-steps.el
```diff
@@ -66,20 +66,24 @@
...
 (Given "^I call the add-missing-libspec callback directly with mock data to import"
-  (lambda ()
-    (cljr--add-missing-libspec "Date" "java.util.Date" "import")))
+       (lambda ()
+         (cljr--add-missing-libspec "Date" '((java.util.Date :class)))))
 
 (Given "^I call the add-missing-libspec callback directly with mock data to refer split"
-  (lambda ()
-    (cljr--add-missing-libspec "split" "clojure.string" "require")))
+       (lambda ()
```
expez: Did you add indentation rules manually to get `Given` to indent in this manner, @magnars? 

`aggressive-indent-mode` ruined this for me and I was too lazy to turn it off to indent manually :p

magnars: No, I just copied those from the docs back in 2012 and never reindented the
buffer, it seems.

PR #121, opened 2015-03-10
Refactor to reflect api changes in refactor-nrepl

expez:
- find-debug-fns and find-symbol are now toplevel ops
- The return value of find-usages has changed

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -535,11 +535,11 @@ word test in it and whether the file lives under the test/ directory."
   (interactive)
   (cljr--assert-middleware)
   (let* ((body (replace-regexp-in-string "\"" "\"" (buffer-substring-no-properties (point-min) (point-max))))
-         (result (cljr--call-middleware-sync "value"
-                                             (list "op" "refactor"
-                                                   "ns-string" body
-                                                   "refactor-fn" "find-debug-fns"
-                                                   "debug-fns" cljr-debug-functions)))
+         (result (cljr--call-middleware-sync
+                  (list "op" "find-debug-fns"
+                        "ns-string" body
+                        "debug-fns" cljr-debug-functions)
+                  "value"))
          (debug-fn-tuples (pop result))
          (removed-lines 0))
     (while debug-fn-tuples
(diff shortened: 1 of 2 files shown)


clj-refactor.el
```diff
@@ -1952,20 +1944,15 @@ root."
...
-                            cljr--read-symbol-metadata))
-           (buffer-of-symbol (cider-find-var-file (concat ns "/" symbol-name)))
-           (tooling-buffer-p (cider--tooling-file-p (buffer-name buffer-of-symbol))))
-      (cljr--rename-symbol ns occurrences new-name)
-      (when (and (not tooling-buffer-p) symbol-name)
-        (message "Reloading buffer %s where symbol is defined" buffer-of-symbol))
-      (message "Renamed %s occurrences of %s" (length occurrences) (or symbol-name symbol))
-      (when (not tooling-buffer-p)
-        (when symbol-name
-          (cider-load-buffer buffer-of-symbol))
```
expez: I removed the loading of buffers @benedekfazekas.  I don't see the point in reloading only the buffer from which the request was sent.  If anything should be done at all every buffer that's been touched should be reloaded.  We might take it upon ourselves to call `cider-refresh` here.  What do you think?

benedekfazekas: I guess the point here is to try to find the balance between reloading stuff so the new symbol is available but doing as few as reloading as possible. You can easily argue that even the first one is not really important as
- further refactorings possible as we do our own `eval` when building the AST
- when the user finishes the whole thing she will reload the namespaces anyway I assume...

clj-refactor.el
```diff
@@ -1887,39 +1907,25 @@ root."
...
-                             (or (plist-get symbol-meta :col-end) (line-end-position))))
+               (col-end (if (= line-beg line-e)
+                            (line-end-position)
+                          (or (plist-get symbol-meta :col-end) (line-end-position))))
                (end (progn (goto-char (point-min))
                            (forward-line (1- line-e))
-                           (move-to-column column-end)
+                           (move-to-column col-end)
                            (point)))
                (orig-name (plist-get symbol-meta :name))
```
expez: I just tried running this branch (can't get a repl going on the computer at work with this branch due to what looks like `mranderson` problems) and `orig-name` here is fully qualified so the renaming fails.  I haven't touched the code in the middleware creating the name, was it always like this @benedekfazekas?

PR #133, opened 2015-04-01
Make easy to discover refactorings

maio:
This is WIP. I would appreciate any feedback.

TODO:
- make tests compatible with run-with-timer


expez:
Thanks! :+1: 

Could we inline everything into `clj-refactor.el`?

I'm thinking we should just enable this by default, so hitting whatever shortcut `cljr` is bound to will open the discover menu, and then add a `defcustom` to get the old behavior back.  What do you guys think, @benedekfazekas, @magnars ?

magnars:
I think we should suggest the discover menu as a good default for new
people in the readme, but let veterans run their mnemonic shortcuts with no
flashing discover menu. :)

PR #135, opened 2015-04-06
Add helm support

philjackson:
Introduces a basic helm source. Makes Helm a hard dependency.


philjackson:
I just saw the other pull request which looks to do almost exactly the same thing. That'll teach me...

expez:
This looks cool!  Any idea thoughts about whether we should pick `helm` over `discover`?

benedekfazekas:
in case this go forward can you pls squash the commits eventually

philjackson:
There's already a cljr-helm package on melpa. Perhaps if you guys just kept `cljr--all-helpers` or similar, then I could just have that package read its candidates from clj-refactor itself. That way people can just install it if they want, and you guys don't have to bear the weight of the Helm dep.

expez:
Sounds like a good idea :+1:

philjackson:
Ok, back down to just the datastructure. Hopefully nothing changes for your users.

philjackson:
I've squashed it.

expez:
The build appears to be failing now

philjackson:
Hurrah for tests. Fixed.

expez:
Thanks! :+1:

PR #141, opened 2015-04-30
[Fix #131] Add cljr-version

### LINKED ISSUE #131: Add a version command
Full version information for both `clj-refactor` and `refactor-nrepl`, down to the commit.

This makes it easier to deal with support requests.

expez:
We're also checking that the client and server versions are in sync and
if not we emit a warning

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -34,6 +34,9 @@
 (require 'clojure-mode)
 (require 'cider)
 
+(defvar cljr-version "1.0.5-SNAPSHOT"
+  "The current version of clojure-refactor")
+
 (defcustom cljr-add-ns-to-blank-clj-files t
   "When true, automatically adds a ns form to new clj files."
   :group 'cljr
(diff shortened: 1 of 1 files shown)


clj-refactor.el
```diff
@@ -149,7 +152,10 @@ with the middleware."
 
 (defvar cljr--nrepl-ops
   '("artifact-list" "artifact-versions" "clean-ns" "configure" "find-symbol"
-    "find-unbound" "hotload-dependency" "resolve-missing"))
+    "find-unbound" "hotload-dependency" "resolve-missing" "version"))
+
+(defvar cljr-version "1.0.5-SNAPSHOT"
```
benedekfazekas: is this a duplicate?

expez: yes, it duplicates the version in the header of `clj-refactor.el`

benedekfazekas: so you need `(defvar clj-version "1.0.5-SNAPSHOT")` twice?! sry for being slow here...

expez: haha, I'm the one being slow here.  I was so annoyed by this duplication, with the version information appearing both in a var and in the header.  I spent like 10 mins trying to find a way to use `eval-when-compile` to extract the info from the header and put it into the var.

After all that I didn't see the literal duplication when you pointed it out.  Fixed in https://github.com/clojure-emacs/clj-refactor.el/commit/b0f1abaae0ea80b4ebeb45ac2f3474e426e050de

benedekfazekas: :) :+1:

Issue #127, opened 2015-03-30
Some assistance for records and protocols


mtnygard:
This is a feature request for some actions that would be useful in working with types.

**Convert reify to record** This would extract a reify block out into a record, replacing the reify function with a call to construct the new record type.

**Reify a protocol or interface by name** This would add the protocol or interface name as well as stubs for the functions that are part of the named interface or protocol.

expez:
Rebasing creates quite a bit of spam but both of these are currently fixed on master.  Let me know if this doesn't solve your problem.

PR #148, opened 2015-05-09
Add cljr-add-stubs

expez:
This command adds a stub implementation for the interface or protocol at
point.

This is the second half of #127 

https://github.com/clojure-emacs/refactor-nrepl/pull/71


arichiardi:
:+1: In ccw I had some (very naive) [function](https://github.com/laurentpetit/ccw/blob/ae2e5174f68b1649f61ab24b1e8cd549ef4ee6b8/ccw.core/src/clj/ccw/interop.clj#L71) for automagically produce a ```reify``` from a class. I was waiting for this to appear here :smile:

expez:
@arichiardi you thinking of moving over to emacs for clojure development  Or perhaps using refactor-nrepl in ccw?

@magnars @benedekfazekas these tests are failing because I opted to depend on emacs 24.4 when writing edn.el.   Do we really want to support these old versions of emacs?  I'm not very motivated to track down bugs in old versions of emacs.

I'm not super happy with the name of this command, bit does match what it does pretty well. 

I'm also not sure if the typehints in the parameter list add any value.  I figured it would be nice to get a reminder, but I suppose just as often you'll be looking at the docs / already know what to do.

arichiardi:
@expez at the moment I am using both ```cider``` (+ goodies) AND ```ccw```, depending on what I am developing, for mixed projects I find ```ccw``` very very good (and worth improving, as ```cursive``` won't be open sourced, no?).

About the second question, I thought about it, but @laurentpetit has a better idea of the milestones of ```ccw```. It would be really great though. He is thinking about it integrating ```nrepl/cider``` features, see [here](https://github.com/laurentpetit/ccw/issues/734).

laurentpetit:
@expez @arichiardi CCW will definitely have refactoring commands implemented. I don't know much yet about the "client" (equivalent to `clj-refactor.el`) part, which would leverage the `framework` available in Eclipse, to be able to say whether `refactor-nrepl` would provide sufficient information or not. Definitely worth working on / checking eventually.

PR #150, opened 2015-05-21
create function stub from example usage

benedekfazekas:
- if argument is a word use it
- if argument is a string, s-expression, hash/set use name template (arg0, arg1 etc)
- documentation, demo gif, changelog


expez:
:+1:

PR #151, opened 2015-05-22
[Fix #101] Add cljr-inline-symbol

### LINKED ISSUE #101: Inlining
I'd like to be able to inline

* let-bound variables
* vars
* functions

A function would turn into the `fn` form and then we can consider adding `cljr-demote-function` to turn an `fn` into a function literal.

expez:
Make it possible to inline:

- def
- defn
- let-bound vars

This is bound to is

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master (unreleased)
 
+- Add `cljr-inline-symbol` which replaces the symbol at point with its definition.
 - add `cljr-add-stubs` which adds a skeleton implementation of the protocol or interface at point.
 - Add `cljr-reify-to-defrecord`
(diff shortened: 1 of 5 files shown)


benedekfazekas:
nice! 

bit of explanation/description/animgif would not hurt tho...

expez:
I did some work on the wiki last week, I really want to get everything up in there, but I haen't found a good way to record gifs on linux :(

benedekfazekas:
i anyway want to try it so will do in the afternoon

arichiardi:
For recording, there is [this](http://askubuntu.com/a/201018) and thanks for having implemented this feature!

expez:
Thanks for the tip @arichiardi!

Issue #155, opened 2015-05-29
Extraneous whitespace after project clean.


philjackson:
Sometimes, when running project clean, the libspecs have whitespace appended to them, even if the libspecs themselves haven't changed:

![extranious-whitespace](https://cloud.githubusercontent.com/assets/43794/7880052/ef7fb0b0-05ed-11e5-9ae6-f83cd2dc4285.png)

benedekfazekas:
what is in your project clean: remove requires or clean ns?

philjackson:
I'd just come back to mention that I only have `cljr-remove-unused-requires` in there. Presumably sort would clean out that WS?

philjackson:
Looking at the code (I probably should have done first) it's an easy enough fix to call `delete-trailing-whitespace` on the region that's already defined. Shall I patch?

expez:
That would be great :+1:

PR #156, opened 2015-05-29
Clean trailing whitespace when removing libspecs.

philjackson:
Issue referenced at https://github.com/clojure-emacs/clj-refactor.el/issues/155


expez:
Would you mind force pushing an updated commit message where the trailing `.` is removed and a prefix of `[Fix #155]` is added?

philjackson:
Done.

expez:
It seems the tests were written with the trailing whitespace in mind

benedekfazekas:
I guess the tests need a fix in this case

expez:
Thanks! :+1:

PR #162, opened 2015-06-04
Teach cljr--add-test-use-declarations how to find source ns

arichiardi:
This improvement/fix is basically what we discuss [in here](https://github.com/bbatsov/projectile/issues/642#issuecomment-107986794). It allows filling up the newly created test buffer with the right source namespace.

The source namespace is not obtained by chopping ```_test``` only, but it is actually looked up in the ```src``` folder.

This is the biggest ```elisp``` piece of code I have ever written (my first baby) and it might containe mistakes which I will fix asap of course in case :smile: 

Together with this addition, [mechanical test generation](https://github.com/bbatsov/projectile/issues/642#issuecomment-107596828) can finally come to life.

@expez bbatsov/projectile#642

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Add `cljr-reify-to-defrecord`
 - Add `cljr-show-changelog` so users don't have to visit github to find out what's changed after a package update.
 - Add `cljr-create-fn-from-example` to create function stub based on example usage
+- Now cljr--add-test-use-declarations actually checks the file system in order to find its require for the source ns.
 
 ## 1.0.5
 
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -560,11 +560,27 @@ to us to make sure it's nicely indented."
     (goto-char (point-min))
     (search-forward package nil t)))
 
+(defun cljr--find-source-ns-of-test-ns (test-ns test-file)
+  (let* ((ns-chunks (split-string test-ns "[.]" t))
+         (test-name (car (last ns-chunks)))
+         (src-dir-name (s-replace "test/" "src/" (file-name-directory test-file)))
+         (replace-underscore (-partial 's-replace "_" "-"))
+         (src-ns (car (--filter (or (s-prefix-p it test-name)
+                                    (s-suffix-p it test-name))
+                                (-map (lambda (file-name)
+                                        (funcall replace-underscore
+                                                 (file-name-sans-extension file-name)))
+                                      (directory-files src-dir-name))))))
+    (when src-ns
+      (mapconcat 'identity (append (butlast ns-chunks) (list src-ns)) "."))))
+
 (defun cljr--add-test-use-declarations ()
   (save-excursion
-    (let ((ns (clojure-find-ns)))
+    (let* ((ns (clojure-find-ns))
+           (source-ns (cljr--find-source-ns-of-test-ns ns (buffer-file-name))))
       (cljr--insert-in-ns ":require")
-      (insert "[" (s-chop-suffix "-test" ns) " :refer :all]")
+      (when source-ns
+        (insert "[" source-ns " :refer :all]"))
       (cljr--insert-in-ns ":require")
       (insert "[" (cond
                    ((cljr--project-depends-on-p "midje") "midje.sweet")
--- a/features/support/env.el
+++ b/features/support/env.el
@@ -33,8 +33,8 @@
  (add-hook 'clojure-mode-hook (lambda () (clj-refactor-mode))))
 
 (Before
- ;; Before each scenario is run
- )
+ (save-all-buffers-dont-ask)
+ (kill-matching-buffers-dont-ask "clj"))
 
 (defun save-all-buffers-dont-ask ()
   (dolist (buffer (buffer-list))
--- a/features/auto-ns.feature
+++ b/features/auto-ns.feature
@@ -2,6 +2,7 @@ Feature: Add namespace to blank .clj files
 
   Background:
     Given I have a project "cljr" in "tmp"
+    When I have a clojure-file "tmp/src/cljr/core.clj"
 
   Scenario: Source file
     When I open file "tmp/src/cljr/core.clj"
@@ -12,6 +13,24 @@ Feature: Add namespace to blank .clj files
     Then I should see:
     """
     (ns cljr.core-test
+      (:require [cljr.core :refer :all]
+                [clojure.test :refer :all]))
+    """
+
+  Scenario: Test file with foo_ prefix
+    When I open file "tmp/test/cljr/foo_core.clj"
+    Then I should see:
+    """
+    (ns cljr.foo-core
+      (:require [cljr.core :refer :all]
+                [clojure.test :refer :all]))
+    """
+
+ Scenario: Test file with _bar suffix
+    When I open file "tmp/test/cljr/core_bar.clj"
+    Then I should see:
+    """
+    (ns cljr.core-bar
       (:require [cljr.core :refer :all]
                 [clojure.test :refer :all]))
     """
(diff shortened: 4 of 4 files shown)


arichiardi:
Will fix the test.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - Add `cljr-reify-to-defrecord`
 - Add `cljr-show-changelog` so users don't have to visit github to find out what's changed after a package update.
 - Add `cljr-create-fn-from-example` to create function stub based on example usage
+- now cljr--add-test-use-declarations actually checks the file system in order to find its require for the source ns.
```
expez: Capitalize this to match the other entries.  You could also do a slightly better job of explaining what's actually changed.  You've added support for various other test prefixes and suffixes instead of relying on the convention of `_test` suffix, right?

clj-refactor.el
```diff
@@ -560,11 +560,28 @@ to us to make sure it's nicely indented."
...
     (search-forward package nil t)))
 
+(defun cljr--find-source-ns-of-test-ns (test-ns test-file)
+  (let* ((ns-chunks (split-string test-ns "[.]" t))
+         (test-name (car (last ns-chunks)))
+         (src-dir-name (s-replace "test/" "src/" (file-name-directory test-file)))
+         (replace-underscore (-partial 's-replace "_" "-"))
+         (src-ns (car (--filter (or (string-prefix-p it test-name)
+                                    (string-suffix-p it test-name))
+                                ;; Cannot find dash.el's -compose?
```
expez: This comment seems unnecessary.  Either solve the problem (are we using an old version of dash?) or proceed with confidence like everything is honky dory :p

expez:
Squash the commits, limit the first line of the commit message to 50 characters (I've added a [contributing.md](CONTRIBUTING.MD), matching the practices of `clojure-mode`, `cider` etc detailing these draconian rules).

You should also add some tests to both make sure this functionality works as intended and to ensure that we don't accidentally break the feature as we stumble around the codebase later on.

clj-refactor.el
arichiardi: I think yes the ```-compose``` fn came later. I will stick with this implementation ok?

expez: yeah, I'm fine with that, but you should drop the comment in any event.

arichiardi:
Ok, will do everything and...yes I am checking the tests, I have just need to add a fake source file because now it is needed in order to pass.

CHANGELOG.md
arichiardi: Yes, it basically supports every kind of "convention" as the code checks if the source file (without dir and extension) is substring of the test ns.

arichiardi:
Is there a way to execute the test individually and stop to debug during execution? I am new to this ```.feature``` files :smile:

arichiardi:
Can't figure out a way to create a temp file, incredible :smiling_imp:

benedekfazekas:
you can run the failing tests only with `--only-failing` (eg. `./run-tests.sh --only-failing`) not sure about the stop and debug. more detailed docs:

https://github.com/ecukes/ecukes

benedekfazekas:
creating a file: i suppose https://github.com/clojure-emacs/clj-refactor.el/blob/master/features/step-definitions/clj-refactor-steps.el should give some hints...

arichiardi:
Yes, I added:

```
    Scenario: Midje
      When I open file "tmp/project.clj" 
      -> And I have a clojure-file "tmp/src/cljr/core.clj"
      And I press "C-M-f"
```

but still some problem, thanks for the info btw...

arichiardi:
Mmm...I have a:

```/.emacs.d/.../submodules/clj-refactor/tmp/src/example/: no such directory```

Why is it creating this file / folder in there?

**Comment by benedekfazekas** [comment]
try something like:
```
    Given I have a project "cljr" in "tmp"
    And I have a clojure-file "tmp/src/cljr/core.clj"
```

**Comment by arichiardi** [comment]
Ok trying, thanks for your help.

**Comment by arichiardi** [comment]
Are the tests executed sequentially or in parallel?

**Comment by benedekfazekas** [comment]
i think sequentially

**Comment by arichiardi** [comment]
BTW there was this at the beginning:

```
Background:
    Given I have a project "cljr" in "tmp"
```

I don't think I need to add another one...am I right?

**Comment by benedekfazekas** [comment]
in that case you should not be needing it again...

**Comment by arichiardi** [comment]
It looks like some autosaving is going on...is it possible in your opinion?...I really don't know where the ```example``` folder comes from...

**Comment by expez** [comment]
@arichiardi how are you running the tests?  If you're using `./run-tests.sh` in the clj-refactor folder then it shouldn't be creating dirs under ~/.emacs.d

**Comment by arichiardi** [comment]
No don't worry...I have a submodule and developing in there...

BTW I added:

```elisp
(Before
 (save-all-buffers-dont-ask)
 (kill-matching-buffers-dont-ask "clj")
 )
```

And now they are running, I see my test failing still but at least is not some strange side effect of some other test, don't know why it was keeping some other buffer open (with ```(ns example.core)``` inside).

arichiardi:
Can't believe after one day I was able to have zero errors..

arichiardi:
This should finish it off :facepunch: 
The ```ecukes``` framework was fun to work with once I fixed the problems.
Some error was swallowed by:
```elips
(defun cljr--add-ns-if-blank-clj-file ()
  (ignore-errors
  ...
```
Some other was just some side effect of previous tests that I solved putting the resetting functions not only in ```(After ...)``` but also in ```(Before ...)```

features/support/env.el
```diff
@@ -18,6 +18,31 @@
 (require 'ert)
 (require 's)
 
+;;;; Compatibility - From Projectile
+(eval-and-compile
+  ;; Added in Emacs 24.3.
+  (unless (fboundp 'defvar-local)
```
expez: We only support `24.3` or newer so we don't need this.

features/support/env.el
```diff
@@ -18,6 +18,31 @@
...
+      "Define VAR as a buffer-local variable with default value VAL.
+Like `defvar' but additionally marks the variable as being automatically
+buffer-local wherever it is set."
+      (declare (debug defvar) (doc-string 3))
+      `(progn
+         (defvar ,var ,val ,docstring)
+         (make-variable-buffer-local ',var))))
+
+  ;; Added in Emacs 24.4
+  (unless (fboundp 'string-suffix-p)
```
expez: This isn't needed either as we're relying on `s` which has `s-suffix?` as well as and alias called `s-suffix-p`.  As an aside, `string-suffix-p`, and a few other new string functions, were written because core emacs developers saw how incredibly useful `s.el` was :)

arichiardi: Weird...It was giving an error exactly on ```string-suffix-p```, that is why I added it.

features/support/env.el
arichiardi: Ok

features/support/env.el
arichiardi: In any case, I will use ```s-suffix-p``` instead if it's the same thing :smile:

expez: Well you would have to use the variants found in `s`, the names aren't identical  :)

expez:
Remember to squash the commits and we should be good to go!

arichiardi:
Done, sorry for the double push, I had to cut some text from the first line, every test runs fine, thanks for the ```s-suffix-p``` hint.

expez:
Thanks! :+1:

arichiardi:
:bow: (strange bow this one)

benedekfazekas:
if it hurts do it more often ;) :+1:

arichiardi:
lol is true, ok I don't wanna know what's going on there...

PR #163, opened 2015-06-04
Fix promote-fn to work with named anonymous fns

luxbock:
- If the fn is named then use that name without asking

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -431,10 +431,16 @@ e.g. `re-search-forward'"
 
 (defun cljr--goto-fn-definition ()
   (let* ((search-bound (cljr--point-after 'cljr--goto-toplevel))
-         (literal (cljr--point-at-text-matching "#(" :backward search-bound
-                                                :noerror))
-         (fn (cljr--point-at-text-matching "(fn \\[" :backward search-bound
-                                           :noerror)))
+         (literal (cljr--point-at-text-matching
+                   "#("
+                   :backward
+                   search-bound
+                   :noerror))
+         (fn (cljr--point-at-text-matching
+              "(fn \\(\\_<[^ ]+\\_>[[:space:]\n]+\\)?\\["
+              :backward
+              search-bound
+              :noerror)))
     (if (or literal fn)
         (progn (goto-char (max (or literal (point-min)) (or fn (point-min))))
                (when (looking-back "#")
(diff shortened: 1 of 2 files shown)


expez:
This is great, thanks!

Would you mind adding a test or two demonstrating/validating the new behavior?

luxbock:
My first time writing tests for Elisp :)

Right now promote-fn doesn't work if it's called from the hash-tag or before the argument vector of an anonymous function, so I think I'll try to fix that as well.

expez:
> Right now promote-fn doesn't work if it's called from the hash-tag or before the argument vector of an anonymous function, so I think I'll try to fix that as well.

Cool!

luxbock:
Hmm think this might still need a bit more work.

luxbock:
I believe this should work now.

luxbock:
Except Travis is failing. I don't get it. The tests pass when I run them locally.

benedekfazekas:
let me rerun the build. not as if i don't trust travis...

benedekfazekas:
nope. it hangs with 24.3. try to run the tests with 24.3

luxbock:
Grr... The problem was that `<` only works with two arguments prior to 24.1. Why have Polish notation at all if `<` can't accept variable number of args!?

clj-refactor.el
```diff
@@ -429,17 +429,31 @@ e.g. `re-search-forward'"
...
-      (error "Can't find definition of anonymous function!"))))
+  (if (zerop (first (paredit-current-parse-state)))
+      (error "Not inside a s-expression.")
+    (let* ((pt-orig (point))
+           (search-bound (cljr--point-after 'cljr--goto-toplevel))
+           found-fn-p)
+      (while (not found-fn-p)
+        (paredit-backward-up)
+        (-if-let (fn-beg (cljr--point-for-anon-function))
+            (let ((fn-end (save-excursion (paredit-forward) (point))))
```
expez: This should probably be `(cljr--point-after 'paredit-forward)`

clj-refactor.el
```diff
@@ -429,17 +429,31 @@ e.g. `re-search-forward'"
...
-        (progn (goto-char (max (or literal (point-min)) (or fn (point-min))))
-               (when (looking-back "#")
-                 (backward-char)))
-      (error "Can't find definition of anonymous function!"))))
+  (if (zerop (first (paredit-current-parse-state)))
+      (error "Not inside a s-expression.")
+    (let* ((pt-orig (point))
+           (search-bound (cljr--point-after 'cljr--goto-toplevel))
+           found-fn-p)
+      (while (not found-fn-p)
```
expez: Won't this loop forever if there's no fn definition between here and the toplevel of the file?  I think maybe this needs to be guarded with a check on `cljr--toplevel-p?`

luxbock: It throws via `(error "Can't find definition of anonymous function!")` if point reaches the top-level, i.e. `search-bound`, which in turn is found via calling `cljr--goto-toplevel`, which also uses `paredit-backward-up`, so I'm fairly sure it should work.

expez:
Thank for addressing these issues!

PR #169, opened 2015-06-15
Add cljr-change-function-signature

expez:
Refactor the function signature of the function at point.  Supports both
changing the order and name of the function's parameters.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## Up next
-
+- Add `cljr-change-function-signature` to re-order or re-name function parameters.
 - Keep pressing `l` after `cljr-expand-let` to expand further.
 
 ## 1.1.0
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -197,6 +197,11 @@ with the middleware."
 
 (defvar cljr--debug-mode nil)
 
+(defvar cljr--occurrences nil)
+(defvar cljr--signature-changes nil)
+(defvar cljr--change-signature-buffer "*cljr-change-signature*")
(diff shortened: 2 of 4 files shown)


expez:
.This feature needs a bunch of tests, but I'd like som feedback at this point.

The user is presented with a buffer containing the function parameters.  In this buffer the user can use `M-n` or `M-p` to move the parameters of the parameter list or `C-c C-e` to change the name of a parameter.  When the user is satisfied with her changes she can press `C-c C-c` to commit and we go off an update all call-sites.  There's also `C-c C-k` to abort.

Most call-sites can be handled automatically (regular calls, in addition to `apply` and `partial`), but for those we can't handle the entry is added to a buffer containing stuff which needs manual intervention.  This buffer is a `grep` buffer much like the buffer for `find-usages` where the user can easily jump around.

benedekfazekas:
nice one! I will try to give it a spin in the evening..

Malabarba:
Sounds like a great feature. Since the signatures buffer is in view mode, would it make sense to also offer unprefixed keys on it? Like q, e, and RET, to respectively quit, edit, and confirm.

expez:
> Sounds like a great feature. Since the signatures buffer is in view mode, would it make sense to also offer unprefixed keys on it? Like q, e, and RET, to respectively quit, edit, and confirm.

Great idea!

expez:
@magnars any thoughts about these test failures?  I haven't touched any code related to the failures and it's all green locally :/

magnars:
Hmm. Seems like `project-clean` and `move-form` are failing. They both use `remove-unused-requires` - has that been changed?

PR #171, opened 2015-06-16
Fix cljr-unwind-all

Malabarba:
Without this, the unwind-all loop never stops and ends up unwinding the etire file

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -1325,12 +1325,15 @@ optionally including those that are declared private."
 
 ;;;###autoload
 (defun cljr-unwind ()
+  "Unwind thread at point or above point by one level.
+Return nil if there are no more levels to unwind."
   (interactive)
   (ignore-errors
     (forward-char 3))
   (search-backward-regexp "([^-]*->")
   (if (cljr--nothing-more-to-unwind)
-      (cljr--pop-out-of-threading)
+      (progn (cljr--pop-out-of-threading)
+             nil)
     (paredit-forward-down)
     (cond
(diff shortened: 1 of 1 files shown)


expez:
If I change the test for `cljr-unwind-all` to be and let `|` indicate point:

```clj
    (-> foo bar baz)

    |(-> {}
        (assoc :key "value")
        (dissoc :lock))
```

It only unwinds the last form.  Is this some regression caused by a newer version of paredit?

The tests are currently run on paredit 24, is that what you're using when you see this behavior?

expez:
This also works when I run it in my own emacs instance, where I seem to have paredit 25 beta (I've switched to smartparens)

Malabarba:
I'm on paredit `25beta`.

I don't know if this is a regression, but in any case, does cljr want to depend on the return value of a function from another package that has no documented return value? When I check the source of `paredit-raise-sexp` the last form is a call to `indent-region`, which also has no documented return value and could vary with Emacs version.

Malabarba:
In my particular case, `pop-out-of-threading` always returns non-nil because the last form in `indent-region` is `(setq deactivate-mark t)`. Maybe you don't have that problem because our Emacs versions are different?  I'm on 25

expez:
I completely agree with everything you say :)  My intention was merely to figure out if something had changed which might mess up other stuff.

expez:
Thanks! :+1:

Malabarba:
> My intention was merely to figure out if something had changed which might mess up other stuff.
Ah well, if I had to guess, I'd say something probably changed in `indent-region`. Whether that messes anything depends on whether cljr relies on paredit return values anywhere else. :)

PR #175, opened 2015-06-23
[Fix #174] make clean-ns comptabile with project clean

### LINKED ISSUE #174: when using clean-ns with project-clean some ` *temp file* files get created
reported by @martinrojer. i was able to reproduce. these temp files are created for in two subdirectories of my project. content seems to be coming from two namespaces of my project (none of the is really changed by the clean)

benedekfazekas:
not entirely happy with this but could not come up with a better solution. any advice @expez @magnars ?

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -2228,10 +2228,18 @@ root."
   (interactive)
   (cljr--assert-leiningen-project)
   (cljr--assert-middleware)
-  (save-buffer)
-  (let ((result (nrepl-send-sync-request
+  ;; don't save the buffer preliminarily if we are called from project clean
+  (when (not (boundp 'filename))
+    (save-buffer))
+  ;; if filename *is* bound `cljr-clean-ns was called from project-clean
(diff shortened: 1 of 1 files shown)


expez:
What problem is this solving?

clj-refactor.el
```diff
@@ -2228,10 +2228,14 @@ root."
   (interactive)
   (cljr--assert-leiningen-project)
   (cljr--assert-middleware)
-  (save-buffer)
-  (let ((result (nrepl-send-sync-request
+  (when (not (boundp 'filename))
+    (save-buffer))
+  (let* ((path-to-file (if (boundp 'filename)
```
expez: I believe this is constantly false.  I can't find a variable named `filename` when I check.

benedekfazekas:
solves #174 

filename is defined in a macro called `cljr--update-file` used by `cljr-project-clean`

benedekfazekas:
to reproduce the problems just configure you project-clean to run clean-ns and run it

PR #179, opened 2015-06-30
Introduce let form when attempting to move-to-let without one

oliyh:
I noticed when using `move-to-let` that, if I wasn't already inside a `let` form, it would attach itself to the top level form - my function declaration - and into the function bindings. This change will instead introduce a new `let` form if it gets to the top level expression without finding one.

Thanks for all the excellent work!


benedekfazekas:
looks cool to me. thx!

magnars:
:+1:

PR #192, opened 2015-07-16
add cljr--looking-back for looking back near (point-min)

lewang:
Repro:

pressed `M-up` near beginning of buffer and error is thrown.

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -379,9 +379,15 @@ if SAVE-EXCURSION is T POINT does not move."
           (search-forward s bound t))
       (search-forward s bound t))))
 
+(defun cljr--looking-back (regexp &optional limit greedy)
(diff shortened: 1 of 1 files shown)


clj-refactor.el
```diff
@@ -379,9 +379,15 @@ if SAVE-EXCURSION is T POINT does not move."
           (search-forward s bound t))
       (search-forward s bound t))))
 
+(defun cljr--looking-back (regexp &optional limit greedy)
+  "Call `looking-back', but tolerate limit > point."
```
expez: Shouldn't this be `limit < point`?

lewang: You're right.  In fact all calls to look-back specifying limit in this file is using it wrong.  

In fact using limit isn't really necessary if we use the \= boundary.   

I'll close this and make another PR.

Thanks @expez

PR #193, opened 2015-07-16
fix issue with LIMIT of looking-back

lewang:
- LIMIT is not how far you want to look back, but the buffer position to
  stop looking.
- Instead of using any LIMIT parameter, "\=" can be used to anchor at
  point.

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -380,7 +380,7 @@ if SAVE-EXCURSION is T POINT does not move."
 
 (defun cljr--goto-toplevel ()
   (paredit-backward-up (cljr--depth-at-point))
-  (when (looking-back "#")
+  (when (looking-back "#\\=")
     (backward-char)))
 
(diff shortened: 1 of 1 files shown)


clj-refactor.el
```diff
@@ -1641,7 +1641,7 @@ This function only does the actual removal."
 
 (defun cljr--find-symbol-at-point ()
   (save-excursion
-    (when (looking-back "\\s_\\|\\sw" 1)
+    (when (looking-back "\\s_\\\\=|\\sw\\=")
```
Malabarba: I think this should be `\\=\\|`, no?

lewang:
@Malabarba Thanks!  I fixed that.  I didn't notice paredit was being "smart" for me.

expez:
Thanks for taking the time to think about how your current issue affected the whole code base.  I really appreciate the thoroughness!

PR #198, opened 2015-07-23
Improve add-require-snippet

Malabarba:


magnars:
Seems like you need to take a look at the tests. Would also help us to see what exactly the improvement entails. :)

Malabarba:
The change adds a default value to field # 4. This value is the last symbol of the namespace you wrote in field # 3.

I'll check the tests soon.

Malabarba:
Hey @magnars , I'm trying to understand this test, so maybe you can help me?
[Here's a line link](https://travis-ci.org/clojure-emacs/clj-refactor.el/jobs/72307071#L2950).

Is it complaining that the requires are in the wrong order? 

If that's the case, I could just invert the order the test expects, but I'm wondering whether that is indicative of a deeper issue (since I don't know how changing the snippet a little would affect the order here).

magnars:
Yay, I can help with this :)

```
Expected
      (ns cljr.core
        (:require [clojure.strings :as s]
                  [clj-time.core :refer :all]))
to be part of:
      (ns cljr.core
        (:require [clj-time.core :refer :all]
                  [clojure.strings :as s]))
```

is, like you said, a complaint that the requires are in the wrong order. And I think I know why - you probably haven't exited the yasnippet yet - after which the sorting happens. Maybe it would help with another `TAB` ?

Malabarba:
Trying it out here, the test already had enough tabs for the sorting to happen. Maybe when it was written the sorting was off by default?
It's strange that this didn't popup before, but looking at the following test shows that the order used on this test was wrong. Go figure. :P

magnars:
That is weird. Thanks for the improvements :)

PR #205, opened 2015-08-13
Fix bug in cljr--resolve-alias

luxbock:
- The old regexp was too greedy.

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -522,7 +522,7 @@ e.g. `re-search-forward'"
   (save-excursion
     (cljr--goto-ns)
     (when (re-search-forward
-           (format ":as\\s-+%s" (regexp-quote alias))
+           (format ":as\\s-+%s\\_>" (regexp-quote alias))
            (cljr--point-after 'paredit-forward)
            :noerror)
       (paredit-backward-up)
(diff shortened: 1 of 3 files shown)


expez:
How did this manifest?  Would you mind adding a test to ensure we don't break it again when we refactor?

luxbock:
I was trying to use the new `cljr-add-stubs` command in a namespace where I had one ns aliased as `s` and `clojure.string` before it as `str`, and it picked up the wrong symbol to use. 

Sure I'll add a test-case a bit later.

clj-refactor.el
```diff
@@ -522,7 +522,7 @@ e.g. `re-search-forward'"
   (save-excursion
     (cljr--goto-ns)
     (when (re-search-forward
-           (format ":as\\s-+%s" (regexp-quote alias))
+           (format ":as\\s-+s\\>" (regexp-quote alias))
```
Malabarba: Shouldn't this be `":as\\s-+%s\\_>"` (note the % and the _)

luxbock: Whops! Fixed.

expez:
Thanks for adding a test!  Squash the commits and we should be good to go :+1:

expez:
Thanks! :+1:

PR #206, opened 2015-08-14
`sn`: Improve cljr-sort-ns

vmfhrmfoaj:
There are no changes actually, but the buffer has been changed.
when I sort the ns which already sorted.

so If not changed actually, clean the modified flag of buffer.

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -859,15 +859,26 @@ word test in it and whether the file lives under the test/ directory."
 See: https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-sort-ns"
   (interactive)
   (save-excursion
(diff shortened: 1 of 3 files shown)


expez:
Cool, thanks! :+1: 

How come you're not using `cn`?  Do you mostly write cljs?

vmfhrmfoaj:
Another solution outside the `cli-refactor`.

https://github.com/vmfhrmfoaj/.emacs.d/blob/master/configs/100-clojure-lang.el#L218-L231

vmfhrmfoaj:
Code that was created a long time before, did not use the `cli-refactor`.
I have some problem using `cn` when I refactoring the code using `cli-refactor`.
I will report the problem.

PR #218, opened 2015-08-27
[#Fix 217] Parameterise auto-ns test declarations

expez:
- Bunch of new defcustoms: `cljr-midje-test-declaration`,
  `clj-expectations-test-declarations`,
  `cljr-cljc-clojure-test-declarations` and
  `cljr-clojure-test-declarations`.
- Default is no longer `:refer :all` anywhere.
- Require `expectations :as e`
- Require `clojure.test :as t`
- Require `midje.sweet :as midje`

`:refer :all` is a generally frowned upon now and it's not available in
cljs so in a world of cljc these are likely to be less and less used
now.

If the user is unhappy with our choice of aliases for their test
namespaces they can override them using the new defcustoms.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Changes
 
+- [#217](https://github.com/clojure-emacs/clj-refactor.el/issues/217)When requiring the test framework in test files stop favoring `:refer :all`.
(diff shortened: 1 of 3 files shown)


expez:
Would love some feedback on this, `:refer :all` is kind of convenient so I can see this being controversial.

PR #232, opened 2015-09-24
Add basic cljs ns forms support

bsteuber:
Hi, I finally started with basic cljs support ^^
I enabled dealing with :require-macros and :use-macros by restricting cljr--search-forward-within-sexp to only succeed when the string is followed by ), ], or space so that e.g. :require doesn't get matched by :require-macros. It feels a bit hacky to me, but it works :D

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -538,13 +538,18 @@ _s_: Refactor related functions
 
 (defun cljr--search-forward-within-sexp (s &optional save-excursion)
   "Searches forward for S in the current sexp.
-
-if SAVE-EXCURSION is T POINT does not move."
-  (let ((bound (save-excursion (forward-list 1) (point))))
-    (if save-excursion
-        (save-excursion
-          (search-forward s bound t))
-      (search-forward s bound t))))
+   S must be followed by a space, ), or ] character.
+   If SAVE-EXCURSION is T POINT does not move."
+  (let ((bound (save-excursion (forward-list 1) (point)))
+        (search-regex (concat s "[] )]")))
+    (cl-flet ((do-search ()
+                         (when (search-forward-regexp search-regex bound t)
+                           (backward-char)
+                           (point))))
+      (if save-excursion
+          (save-excursion
+            (do-search))
+        (do-search)))))
 
 (defun cljr--goto-toplevel ()
   (paredit-backward-up (cljr--depth-at-point))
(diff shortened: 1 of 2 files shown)


clj-refactor.el
```diff
@@ -538,13 +538,18 @@ _s_: Refactor related functions
...
 (defun cljr--search-forward-within-sexp (s &optional save-excursion)
   "Searches forward for S in the current sexp.
-
-if SAVE-EXCURSION is T POINT does not move."
-  (let ((bound (save-excursion (forward-list 1) (point))))
-    (if save-excursion
-        (save-excursion
-          (search-forward s bound t))
-      (search-forward s bound t))))
+   S must be followed by a space, ), or ] character.
```
Malabarba: Why? Is it meant to detect the end of a symbol? If so `\\_>` is more appropriate.

expez:
This looks good, but I don't like the changes to `cljr--search-forward-within-sexp` I agree that this is a total hack and I don't want to merge it in its present state.

That being, said you might rethink spending any more time on this at all. `clean-ns` now has cljs support.  Rudimentary support was added in `1.2.0-SNAPSHOT`, but a proper support will arrive in 2.0.0-SNAPSHOT which I'll get out this weekend.  In any event, I'd love your help flushing out any bugs in `clean-ns` when it's used on cljs and cljc files as I write mostly clj myself.

bsteuber:
I see...
Well I'd love to try clean-ns, but unfortunately it depends on cider-0.10
which I couldn't get to run at this point. So I'll wait a bit until the
newest cider stabilizes..

Malabarba:
If you let us know what was the problem you had it'll get fixed a lot faster. 😉

bsteuber:
Alright, see https://github.com/clojure-emacs/cider/issues/1349
And thanks for the quick responses :-)

Artur Malabarba <notifications@github.com> schrieb am Do., 24. Sep. 2015
20:12:

expez:
If you don't want expend any more effort on this we should close it, but there might be others in your situations, or a few people who don't run our middleware which might appreciate the effort.

You could also solve your original problem by just changing the regexp for the `:require` close to be `:require[^-]` I think, that way `:require-macros` isn't a partial match anymore.

bsteuber:
Well it's a little harder than that since switching the search to using
regexes will break searches for stuff with parens inside. Now that I think
of it, why the hell did my original fix work? :)

PR #261, opened 2015-11-23
[Fix #235] Remove functions replaced by clean-ns

### LINKED ISSUE #235: Get rid of sort-ns, replace-use and remove-unused-requires in 2.0.0?
Right now a huge chunk of our code base is dedicated to doing work on the ns form.  Part of my motivation for writing `clean-ns` was that these functions were both fragile and difficult to maintain/extend.

Once we solve, the trivial, https://github.com/clojure-emacs/refactor-nrepl/issues/119 there's no technical reason to keep this duplication around.

Does anyone have an objection to removing them?

expez:
- Remove `cljr-remove-unused-requires`.
- Remove `cljr-replace-use`.
- Remove `cljr-sort-ns`.
- Make `cljr-clean-ns` the default function run by `cljr-project-clean`.


Malabarba:
> +14 −1,307

Dear lord. :fearful:

expez:
There's only like 350 lines of affected code, the rest of the changes are tests etc.

I was hoping this would trim down clj-refactor to a larger degree tbh.  We still have 4k lines of elisp left :astonished:

benedekfazekas:
nice one! I guess the only functionality we loose is configurable ns sorting but I expect (1) it is not used  by many (2) easy to reimplement in clojure. perhaps creating some low hanging fruit labelled issues when this is merged still makes sense

expez:
@benedekfazekas we're not losing that.  I implemented the ability to not prune (only tidy up) the ns form a while back using `clean-ns`.

benedekfazekas:
what i meant are the different ways of sorting (alphabetical, by length and semantic (eg. namespaces from the same project on top)) or you reimplemented that too?

expez:
> what i meant are the different ways of sorting (alphabetical, by length and semantic (eg. namespaces from the same project on top)) or you reimplemented that too?

No, this is indeed missing.

magnars:
:+1: nice work!

PR #268, opened 2015-11-28
Properly return other build tools' project files if project.clj is missing

alexander-yakushev:
`cljr--project-file` always returned `project.clj` if `pom.xml` or `boot.clj` was present.

Also, change `boot.clj` to `build.boot` because AFAIK the latter is the default one.

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -794,16 +794,18 @@ A new record is created to define this constructor."
          (locate-dominating-file default-directory "project.clj")))
       (ignore-errors
         (file-truename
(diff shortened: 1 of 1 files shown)


expez:
Great, thanks a lot @alexander-yakushev!

bbatsov:
Don't we have similar functionality in `clojure-mode`? Seems odd to duplicate it here.

expez:
> Don't we have similar functionality in `clojure-mode`? 

Not to my knowledge.  clj-refactor is the only tool which cares about such a project file AFAIK.

We have something similar, a function to return the project root dir, which does look for one of these files.

Issue #157, opened 2015-05-29
rename-symbol only works on the last reference on a line


triclops200:
When using rename-symbol, it only renames the last reference to that symbol on a line.

example renaming size to screen-size:
before:
```clojure
(setup cell-count size [[0 0] [size size]])
```
after:
 ```clojure
(setup cell-count size [[0 0] [size screen-size]])
```
expected:
```clojure
(setup cell-count screen-size [[0 0] [screen-size screen-size]])
```

expez:
Thanks a lot for the bug report!

PR #276, opened 2015-12-26
Replace refer all

benedekfazekas:
Replace :refer :all style require with alias style require or
optionally (with prefix) referred names style require.

- with prefix cljr--replace-refer-all-with-referred-list still available
- add keybinding for cljr-replace-refer-all
- add cljr-replace-refer-all to the ns hydras
- add warm-macro-occurrences-cache
- defcustom to warm-macro-occurrences-cache at startup

clojure-emacs/refactor-nrepl#135 provides the middleware backend function

Intended to be merged for version 2.1.0

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -149,6 +149,12 @@ as can be."
   :group 'cljr
   :type 'boolean)
 
+(defcustom cljr-eagerly-cache-macro-occurrences-on-startup t
+  "If t, the middleware will eagerly populate the macro occurrences cache.
+This makes `cljr-replace-refer-all' as snappy as it can be."
+  :group 'cljr
+  :type 'boolean)
+
 (defcustom cljr-suppress-middleware-warnings nil
   "If t, no middleware warnings are printed to the repl."
   :group 'cljr
@@ -249,8 +255,10 @@ won't run if there is a broken namespace in the project."
     "rename-file-or-dir"
     "resolve-missing"
     "stubs-for-interface"
+    "find-used-publics"
     "version"
     "warm-ast-cache"
+    "warm-macro-occurrences-cache"
     ))
 
 (defvar cljr--debug-mode nil)
@@ -344,6 +352,7 @@ Otherwise open the file and do the changes non-interactively."
     ("ml" . (cljr-move-to-let "Move to let" ?m ("code")))
     ("pc" . (cljr-project-clean "Project clean" ?c ("project")))
     ("pf" . (cljr-promote-function "Promote function" ?p ("code" "toplevel-form")))
+    ("ra" . (cljr-replace-refer-all "Replace refer all" ?a ("ns")))
     ("rf" . (cljr-rename-file-or-dir "Rename file-or-dir" ?r ("project" "toplevel-form")))
     ("rl" . (cljr-remove-let "Remove let" ?r ("code")))
     ("rm" . (cljr-require-macro "Add to or extend the require-macros form" ?M ("ns")))
@@ -402,12 +411,13 @@ Otherwise open the file and do the changes non-interactively."
 ------------------------------------------------------------------------------------------------------------------------------------------------------
 _ai_: Add import to ns                             _am_: Add missing libspec                          _ap_: Add project dependency
 _ar_: Add require to ns                            _au_: Add use to ns                                _cn_: Clean ns
-_rm_: Require a macro into the ns                  _sr_: Stop referring
+_rm_: Require a macro into the ns                  _sr_: Stop referring                               _ra_: Replace refer all
 "
   ("ai" cljr-add-import-to-ns) ("am" cljr-add-missing-libspec)
   ("ap" cljr-add-project-dependency) ("ar" cljr-add-require-to-ns)
   ("au" cljr-add-use-to-ns) ("cn" cljr-clean-ns)
   ("rm" cljr-require-macro) ("sr" cljr-stop-referring)
+  ("ra" cljr-replace-refer-all)
   ("q" nil "quit"))
 
 (defhydra hydra-cljr-code-menu (:color pink :hint nil)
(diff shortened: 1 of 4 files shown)


expez:
We already have `cljr-stop-referring`, but it only works on aliased stuff and doesn't handle `:refer :all`.  Perhaps this new code should just extend `cljr-stop-referring` instead of being its own command?

clj-refactor.el
```diff
@@ -158,6 +158,12 @@ as can be."
   :group 'cljr
   :type 'boolean)
 
+(defcustom cljr-eagerly-cache-macro-occurrences-on-startup t
+  "If t, the middleware will eagerly populate the macro occurrences cache.
+This makes `cljr-replace-refer-all' as it can be."
```
expez: Missing word here

clj-refactor.el
```diff
@@ -1372,6 +1381,38 @@ returns (used.ns.lib1 used.ns.lib2)"
...
+  (when (search-forward ":require" nil t)
+    (let (nses
+          (require-end (save-excursion
+                         (progn
+                           (paredit-backward-up)
+                           (paredit-forward)
+                           (point)))))
+      (if (not (save-excursion (re-search-forward ":refer\s:all" require-end t)))
+          (message "There is no :refer :all in the ns declaration.")
+        (while (save-excursion (re-search-forward ":refer\s:all" require-end t))
```
expez: re-search-forward both moves point and returns a truthy value when it succeeds, so this while loop can be simplified slightly.

clj-refactor.el
```diff
@@ -1372,6 +1381,38 @@ returns (used.ns.lib1 used.ns.lib2)"
...
+          (re-search-forward ":refer\s:all" require-end t)
+          (paredit-backward-up)
+          (let ((prefix (save-excursion
+                          (when (progn
+                                  (paredit-backward-up)
+                                  (not (looking-at "(:require")))
```
expez: prefer `looking-at-p` here

clj-refactor.el
```diff
@@ -2981,8 +3028,8 @@ See: https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-find-usages"
         (when (looking-at-p "(\\s-*def")
           (re-search-forward name)
           (paredit-backward))
-        (when (looking-at (format "\\(.+/\\)?\\(%s\\)" name))
-          (replace-match (format "\\1%s" new-name))))
+        (when (re-search-forward (format "\\(.+/\\)?\\(%s\\)" name) search-limit t)
```
expez: This is what we had back in the day, but we had to change it to the current version because it caused problems.  I think this might introduce a regression.  You should walk the git history to find out exactly why we changed it the last time.

clj-refactor.el
```diff
@@ -3017,6 +3069,51 @@ See: https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-rename-symbol"
...
+                         (--map (gethash :name it) publics-occurrences)))))))
+
+(defun cljr--replace-refer-all-with-alias (ns publics-occurrences alias)
+  (let ((ns-last-token (-last-item (s-split "\\." ns))))
+    (when (re-search-forward (format "\\(%s\s\\):refer\s:all" ns-last-token) nil t)
+      (replace-match (format "\\1:as %s" alias)))
+    (cljr--rename-occurrences ns publics-occurrences (lambda (old-name) (concat alias "/" old-name)))))
+
+;;;###autoload
+(defun cljr-replace-refer-all (with-referred-list)
```
expez: Perhaps a better name here would `replace-referred-symbols?`

clj-refactor.el
```diff
@@ -3017,6 +3069,51 @@ See: https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-rename-symbol"
...
+  "Replaces :refer :all style require with alias :as style require.
+    Also adds the alias prefix to all occurrences of public symbols
```
expez: Indentation here looks a bit weird

features/replace-refer-all.feature
```diff
@@ -0,0 +1,52 @@
+Feature: replace refer all with referred list or aliases
+
+  Background:
+    Given I switch project-clean-prompt off
```
expez: This seems irrelevant

clj-refactor.el
benedekfazekas: hm... I think what you mean is #157 as renaming `arg` in this works fine:

```clj
(defn test [arg]
  (println arg arg arg))
 ```

I guess there is no regression here.

Comment (expez) [inline clj-refactor.el]: I think this might manifest when you're trying to rename `foo` when `foobar` also exists earlier on the same line.

Comment (benedekfazekas) [inline clj-refactor.el]: tested that as well, renaming foo works fine as well:

```clj
(defn testfoo [foobar foo]
  (println foobar foo))
```

hit me with some failing testcase if you still think there is a regression here pls.

expez: > tested that as well, renaming foo works fine as well

:+1:

expez:
I still think we should've extended `cljr-stop-referring` like I suggested above.  Why make it more difficult on ourselves and our users?  I'm already having a hard time remembering all the keybindings and I use this package every day.

clj-refactor.el
```diff
@@ -149,6 +149,12 @@ as can be."
   :group 'cljr
   :type 'boolean)
 
+(defcustom cljr-eagerly-cache-macro-occurrences-on-startup t
```
expez: Why make this a defcustom?  I don't see any reason to ever turn this off.  Do you?

benedekfazekas:
> I still think we should've extended `cljr-stop-referring`

I am not against it. will deffo (re)consider

clj-refactor.el
benedekfazekas: can take a long time and eats quite a bit of CPU at REPL startup. some users might not want this for some reason...

PR #282, opened 2016-01-17
[Fix #266] Add prefix variant to `ar`, `ai` and `au`

### LINKED ISSUE #266: Cljc support for `ar` `au` and `ai`
I think the default should be to modify the clj part and then with a prefix we modify the cljs part of the namespace.

expez:
The prefix variant acts on the cljs part of the cljc ns.  The non-prefix
version works like before and just acts on whatever the first search hit
for matches.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Up next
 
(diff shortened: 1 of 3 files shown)


benedekfazekas:
:+1:

PR #283, opened 2016-01-17
[Fix #267] Add require-macros on `am`

### LINKED ISSUE #267: Analogue to `ar` to require a macro in cljs
I immediately thought of <kbd>am</kbd> for `add-macro`, but this binding is taken by `cljr-add-missing-libspec`.  Perhaps we should move the latter to <kbd>al</kbd>, in 2.0?

expez:
- Rename `cljr-add-missing-libspec` to `cljr-add-libspec`
- Move `cljr-add-libspec` from `am` to `al`.
- Add an equivalent of `cljr-add-require-to-ns` called
  `cljr-add-require-macros-ns` with keybinding `am`.

I think `al` is about as easy to type as `am` so even though adding
libspecs is probably going to see more usage, I don't think this will
change the user's efficiency any, and it keeps our mnemonic shortcust
consistent.


expez:
@magnars @benedekfazekas @Malabarba This change is somewhat controversial, but I think it's for the better.  Furthermore, if we're going to do it, this is the time, as we're now bumping major versions.

benedekfazekas:
hm... `rm` for require macros is not taken tho. I would be reluctant to change a (perhaps) frequently used key combo for a new, perhaps more marginal feature.

expez:
Aight, I'll move the keybindings back before merging.

benedekfazekas:
thx :+1:

Issue #286, opened 2016-01-21
`cljr-clean-ns` with vagrant


timothyandrew:
I use Vagrant for regular development, along with Emacs on OS X.
When I run `M-x cljr-clean-ns` while connected to an nREPL server on the Vagrant VM, I get an error like:

```java
java.io.FileNotFoundException: /Users/tim/work/DummyProject/src/dummy_project/core.clj (No such file or directory)
 at java.io.FileInputStream.open0 (FileInputStream.java:-2)
    java.io.FileInputStream.open (FileInputStream.java:195)
    java.io.FileInputStream.<init> (FileInputStream.java:138)
    java.io.FileInputStream.<init> (FileInputStream.java:93)
    java.io.FileReader.<init> (FileReader.java:58)
    sun.reflect.GeneratedConstructorAccessor14.newInstance (:-1)
    sun.reflect.DelegatingConstructorAccessorImpl.newInstance (DelegatingConstructorAccessorImpl.java:45)
    java.lang.reflect.Constructor.newInstance (Constructor.java:422)
    clojure.lang.Reflector.invokeConstructor (Reflector.java:180)
    refactor_nrepl.core$read_ns_form.invoke (core.clj:202)
    refactor_nrepl.ns.clean_ns$clean_ns.invoke (clean_ns.clj:45)
    refactor_nrepl.middleware$clean_ns_reply.invoke (middleware.clj:75)
    refactor_nrepl.middleware$wrap_refactor$fn__34944.invoke (middleware.clj:123)
    clojure.tools.nrepl.middleware$wrap_conj_descriptor$fn__706.invoke (middleware.clj:22)
    clojure.tools.nrepl.server$handle_STAR_.invoke (server.clj:19)
    clojure.tools.nrepl.server$handle$fn__1105.invoke (server.clj:28)
    clojure.core$binding_conveyor_fn$fn__4444.invoke (core.clj:1916)
    clojure.lang.AFn.call (AFn.java:18)
    java.util.concurrent.FutureTask.run (FutureTask.java:266)
    java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1142)
    java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:617)
    java.lang.Thread.run (Thread.java:745)
```

This is presumably because it's trying to look for the OS X file path (`/Users/tim/...`) on the Vagrant VM, and failing.

[...]

PR #287, opened 2016-01-22
Send the relative buffer path for `clean-ns`

timothyandrew:
- This allows `cljr-clean-ns` to work for remote nREPLs, such as Vagrant.
- This is a fix for #286

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -2774,13 +2774,14 @@ removed."
   ;; Don't save prematurely when called from `cljr-project-clean'
   (unless *cljr--noninteractive*
     (save-buffer))
-  (-when-let (new-ns (cljr--call-middleware-sync
-                      (cljr--create-msg "clean-ns"
-                                        "path" (or path (buffer-file-name))
-                                        "prune-ns-form" (if no-prune? "false"
-                                                          "true"))
-                      "ns"))
-    (cljr--replace-ns new-ns)))
+  (let ((path (or path (cljr--project-relative-path (buffer-file-name)))))
+    (-when-let (new-ns (cljr--call-middleware-sync
+                        (cljr--create-msg "clean-ns"
+                                          "path" path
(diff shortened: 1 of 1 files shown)


timothyandrew:
@expez @benedekfazekas 

This seems to fix the issue. I'd like to hear your thoughts.
Also, since all the features are mocking out the interaction with `refactor-nrepl`, I'm not sure if there's an easy way to test this.

expez:
I'm OK with this if it solves your problem.  I can't think of any instances were a repl is started without having the project root as the working dir.

I'll apply the patch manually, before merging, but i suspect this is going to work.

In addition to this change, the readme of refactor-nrepl has to be updated.  It currently states that an absolute path is required.

Thanks for looking into this, @timothyandrew.

timothyandrew:
> I can't think of any instances were a repl is started without having the project root as the working dir.

The only case I can think of is my particular case: if you're developing on a VM, while writing code outside it. I start the nREPL server inside Vagrant (so the path in there is `/home/vagrant/DummyApp/src/dummy_app/core.clj`), while my emacs is running outside the VM, and the path it sends to nREPL is something like `/Users/tim/code/DummyApp/src/dummy_app/core.clj`, which doesn't exist inside the VM.

> In addition to this change, the readme of refactor-nrepl has to be updated. It currently states that an absolute path is required. 

Sounds good. I'll get on this right away.

> Thanks for looking into this, @timothyandrew.

No worries. Thanks for the quick responses. :+1:

expez:
Aight, this seems to work fine, so i'm good with merging this.  Before I do that, could you prefix the commit message with `[Fix #286]` to match our issue reference convention?

timothyandrew:
@expez Sweet, thanks. I've changed the commit message.

expez:
Thanks @timothyandrew! :+1:

PR #310, opened 2016-03-20
[Fix #306] Fix yasnippet breaking if not connected

### LINKED ISSUE #306: Add-require doesn't jump back
As discussed on gitter, the call to `-sort-ns` errors if there's no connection or if the server doesn't have the refactor-nrepl plugin. Because of that, yasnippet never gets to the point of calling `cljr--pop-tmp-marker-after-yasnippet-1`.

benedekfazekas:
to a REPL when working with the ns form due to trying to sort the ns
afterwards


Malabarba:
Thanks! :-)

benedekfazekas:
no worries :)

PR #311, opened 2016-03-20
[Fix #305] Don't call lookup-alias for keywords

### LINKED ISSUE #305: cljr-slash triggers within keywords
I write lots of Datomic queries which contain keywords with slashes in them (e.g. `:artist/name`), and I wish cljr-slash didn't trigger each time I wrote one as it interrupts for a noticeable instant my typing flow.

I see that cljr uses `cider-in-comment-p` and `cider-in-string-p` to decide on whether it must trigger. I wonder if one could add a check for keywords as well?

benedekfazekas:


clj-refactor.el
```diff
@@ -2199,6 +2199,12 @@ the alias in the project."
           (-when-let (long (cljr--aget cljr-magic-require-namespaces short))
             (list short (list long))))))))
 
+(defun cljr--in-keyword-p ()
+  (cljr--is-keyword?
+   (buffer-substring-no-properties
```
expez: The argument here is just `cider-symbol-at-point`.

Since we already have `cljr--is-keyword?` perhaps we should make `s` optional and act on the symbol at point otherwise?  I also think we should rename `cljr--is-keyword` ro just `cljr--keyword?` so it makes what we have everywhere else.

clj-refactor.el
```diff
@@ -2211,6 +2217,7 @@ form."
   (-when-let (aliases (and cljr-magic-requires
                            (not (cider-in-comment-p))
                            (not (cider-in-string-p))
+                           (not (cljr--in-keyword-p))
```
expez: As @Malabarba pointed out, this would now do the wrong thing if you write `::some-alias/some-keyword`

benedekfazekas: that is true. however, most likely the code in current master does the wrong thing too (have not tested yet) as it does not strip `::` from the prefix so it will never find it in the candidate list. this should be fixed too.

so in summary:

clj-refactor.el
```diff
@@ -2199,6 +2200,12 @@ the alias in the project."
           (-when-let (long (cljr--aget cljr-magic-require-namespaces short))
             (list short (list long))))))))
 
+(defun cljr--in-sans-ns-keyword-p ()
```
expez: I really dislike this name.  I think a better name would be `cljr--in-keyword-sans-alias?`.  We're using `?` for predicates everywhere else, so we should be consistent.

clj-refactor.el
```diff
@@ -2184,9 +2184,10 @@ FEATURE is either :clj or :cljs."
 (defun cljr--magic-requires-lookup-alias ()
   "Return (alias (ns.candidate1 ns.candidate1)) if we recognize
 the alias in the project."
-  (let ((short (buffer-substring-no-properties
-                (cljr--point-after 'paredit-backward)
-                (1- (point)))))
+  (let ((short (->> (buffer-substring-no-properties
```
expez: The `buffer-substring-no-propetrties..` is just `cider-symbol-at-point`.

clj-refactor.el
```diff
@@ -2199,6 +2200,12 @@ the alias in the project."
           (-when-let (long (cljr--aget cljr-magic-require-namespaces short))
             (list short (list long))))))))
 
+(defun cljr--in-sans-ns-keyword-p ()
+  (cljr--sans-ns-keyword?
+   (buffer-substring-no-properties
```
expez: Same here, the thing extracted from the buffer is the same as `cider-symbol-at-point` would return.

clj-refactor.el
benedekfazekas: i am not married to this name. `cljr--in-keyword-sans-alias?` it is then.

apparently it is consistent with the cider naming (already ;) ):
```
(not (cider-in-comment-p))
(not (cider-in-string-p))
```

not married to this either tho. happy to rename if you insist.

clj-refactor.el
benedekfazekas: will check `cider-symbol-at-point` impl and refactor.

clj-refactor.el
```diff
@@ -3653,7 +3661,7 @@ See: https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-create-fn-from-e
         (buffer-substring (point-min) (point-max)))
     s))
 
-(defun cljr--is-keyword? (s)
+(defun cljr--sans-ns-keyword? (s)
```
expez: This function should be `cljr--keyword?` because that's what we care about everywhere else.

You're right that the function is flawed, because it doesn't deal with `::aliased/keywords`, but the intention everywhere else is to just check for keywords, even though the implementation is slightly wrong atm.

I suggest we rename this back and just deal with the ns-checking in `cljr--in-keyword-sans-alias?`.  `cljr--symbol-prefix` can be used to check for an alias.

benedekfazekas: I am reluctant to do that as the new name expresses what the function does better than the original name. no misunderstandings no subtle bugs.

expez: I agree that the name better explains what the function does, but that's only because you found a bug in the function!  The intent everywhere else (than the new usage site added in this PR) is to check for keywords, aliased, fully-qualified, or not.

benedekfazekas: I see your point.

> The intent everywhere else (than the new usage site added in this PR) is to check for keywords, aliased, fully-qualified, or not.

this is an assumption tho. have you checked all call sites if they break when the bug is fixed? I am happy to check and refactor this as you suggesting outside this PR.

expez: > this is an assumption tho. have you checked all call sites if they break when the bug is fixed? 

Yes, I've checked.  Nothing will break :)

benedekfazekas: haha fair enough. will change it back then.

clj-refactor.el
bbatsov: This function should probably have some docstring.

As for predicates - you should really adopt the Emacs conventions as you're writing Emacs Lisp, not Clojure. @magnars is one of the few Elisp hackers who uses `?`, but in `dash.el` even he added the traditional names as aliases. I profoundly dislike violating established conventions for no good reason.

benedekfazekas: ok I rather prefer the `-p` style preds. But really don't want to go down the rabbit hole of renaming all our `foo?` style preds in this PR.
- will create a separate PR replacing all our own `?` predicates with `-p` ones
- keep this fn `-p` style

bbatsov: Fine by me. I was merely pointing out a related observation.

On 25 March 2016 at 20:49, Benedek Fazekas <notifications@github.com> wrote:

> In clj-refactor.el
> <https://github.com/clojure-emacs/clj-refactor.el/pull/311#discussion_r57476657>
> :
>
> > @@ -2199,6 +2200,12 @@ the alias in the project."
> >            (-when-let (long (cljr--aget cljr-magic-require-namespaces short))
> >              (list short (list long))))))))
> >
> > +(defun cljr--in-sans-ns-keyword-p ()
>
> ok I rather prefer the -p style preds. But really don't want to go down
> the rabbit hole of renaming all our foo? style preds in this PR.
>
>    - will create a separate PR replacing all our own ? predicates with -p
>    ones
>    - keep this fn -p style
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> <https://github.com/clojure-emacs/clj-refactor.el/pull/311/files/5f2adca5f6175d34508f83ea951cb48356fbb4c0#r57476657>
>



-- 
Best Regards,
Bozhidar Batsov

http://www.batsov.com

clj-refactor.el
benedekfazekas: `(cider-symbol-at-point)` works here but does not in `(cljr--magic-requires-lookup-alias)`

clj-refactor.el
benedekfazekas: nope, does not work here as `short` needs to be stripped of the `/`. using `(cider-symbol-at-point)` and then stripping the `/` manually does not feel right.

clj-refactor.el
benedekfazekas: thanks for doing that @bbatsov :)

expez:
This looks good now! :+1:

PR #315, opened 2016-03-24
cljr-add-ns-to-blank-clj-files: use cider-expected-ns if available

aiba:
(See clojure-emacs/clojure-mode#372 and clojure-emacs/cider#1622 for context.)

This is the "simple dirty fix" until we migrate the namespace insertion code from clj-refactor to cider.

A side benefit of having both the `clojure-expected-ns` (filesystem heuristic for guessing the namespace) and `cider-expected-ns` (searching the classpath), is that we can use the former when cider is not connected.

I didn't add any tests for the new functionality becuase (a) it's a temp fix and (b) I don't see any tests for clj-refactor.el (are they hiding somewhere nonobvious?).  In any case, I tested it locally with emacs-24.5.1, cider-20160323.1057, and clojure-mode-20160314.2147.

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -1058,8 +1058,7 @@ word test in it and whether the file lives under the test/ directory."
     (when (and cljr-add-ns-to-blank-clj-files
                (cljr--clojure-ish-filename-p (buffer-file-name))
                (= (point-min) (point-max)))
-      (clojure-insert-ns-form)
-      (newline 2)
+      (insert (format "(ns %s)\n\n" (cider-expected-ns)))
       (when (cljr--in-tests-p)
         (cljr--add-test-declarations)))))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Up next
 
+- When inserting ns form to blank clojure-ish file, check if cider is available and connected for better detecting the expected namespace.
+
 ## 2.2.0
 
 - Smarten up `cljr-stop-referring` to replace `:refer :all` style require with alias and apply the alias to all occurrences of symbols from the referred namespace.


clj-refactor.el
```diff
@@ -1058,10 +1058,15 @@ word test in it and whether the file lives under the test/ directory."
     (when (and cljr-add-ns-to-blank-clj-files
                (cljr--clojure-ish-filename-p (buffer-file-name))
                (= (point-min) (point-max)))
-      (clojure-insert-ns-form)
-      (newline 2)
-      (when (cljr--in-tests-p)
-        (cljr--add-test-declarations)))))
+      (let* ((ns (if (and (fboundp #'cider-expected-ns)
+                          (cider-connected-p))
+                     (cider-expected-ns)
```
Malabarba: It would probably be more appropriate for `cider-expected-ns` itself to return `clojure-expected-ns` when not connected.

clj-refactor.el
```diff
@@ -1058,10 +1058,15 @@ word test in it and whether the file lives under the test/ directory."
...
+                   (clojure-expected-ns))))
+        (interactive)
```
Malabarba: This looks out of place.

clj-refactor.el
```diff
@@ -1058,10 +1058,15 @@ word test in it and whether the file lives under the test/ directory."
...
+        (insert (format "(ns %s)" ns))
+        (newline 2)
```
Malabarba: I know this line was already here before, but now is a good time to fix it. This `(newline 2)` can be removed if you change the format string above to `"(ns %s)\n\n"`.

clj-refactor.el
aiba: Yes great point, will change.

clj-refactor.el
aiba: Gotchya, I see (interactive) usually goes at the beginning of the function.  Will fix.

clj-refactor.el
aiba: Good idea.

aiba:
I made the suggested changes.  Side note, by squashing the new commits per the contributor guidelines, github hides the comments as "outdated diff".  If there's a better way to rebase/squash, please let me know.

clj-refactor.el
```diff
@@ -1054,14 +1054,18 @@ word test in it and whether the file lives under the test/ directory."
       (s-ends-with? ".cljc" (buffer-file-name))))
 
 (defun cljr--add-ns-if-blank-clj-file ()
+  (interactive)
```
bbatsov: A private function shouldn't be interactive.

clj-refactor.el
```diff
@@ -1054,14 +1054,18 @@ word test in it and whether the file lives under the test/ directory."
...
   (ignore-errors
     (when (and cljr-add-ns-to-blank-clj-files
                (cljr--clojure-ish-filename-p (buffer-file-name))
                (= (point-min) (point-max)))
-      (clojure-insert-ns-form)
-      (newline 2)
-      (when (cljr--in-tests-p)
-        (cljr--add-test-declarations)))))
+      ;; TODO: replace (if ...) with (cider-expected-ns) once dependable
+      (let* ((ns (if (fboundp #'cider-expected-ns)
```
bbatsov: No need for a `let*` here - there's just a single binding.

benedekfazekas:
Nice one! Pls hold on merging this even if ready until I sort out bumping the versions today both on refactor-nrepl and on cljr

Malabarba:
> by squashing the new commits [...], github hides the comments as "outdated diff". If there's a better way to rebase/squash, please let me know.

No, that's fine. I even find it convenient. :-)

clj-refactor.el
Malabarba: Yes, sorry I was clear @aiba. What I meant before was that you don't need this `(interactive)` at all.

clj-refactor.el
expez: `cider-expected-ns` uses the repl to inspect the classpath right?  Does it fall back to `clojure-expected-ns` if unconnected?  If not we should also check of `cider-connected-p`.

bbatsov: Right now it doesn't, but I'll change this right away.

bbatsov: Done.

aiba:
Thanks for the feedback!  I fixed the two requests and rebased again.

clj-refactor.el
```diff
@@ -1058,10 +1058,13 @@ word test in it and whether the file lives under the test/ directory."
     (when (and cljr-add-ns-to-blank-clj-files
                (cljr--clojure-ish-filename-p (buffer-file-name))
                (= (point-min) (point-max)))
-      (clojure-insert-ns-form)
-      (newline 2)
-      (when (cljr--in-tests-p)
-        (cljr--add-test-declarations)))))
+      ;; TODO: replace (if ...) with (cider-expected-ns) once dependable
```
expez: If `cider-expected-ns` now falls back to `clojure-expected-ns` this `if`, and the todo, can be removed.  Our next release will be right after cider 0.11, I think.

aiba: Oh I see.  Originally I was maintaining compatibility with cider 0.11, but now that we're past cljr 2.2.0 release, I guess we can assume `(fboundp #'cider-expected-ns)`.  Will make this change now...

aiba:
OK, I removed the fallback to `clojure-expected-ns`, since `cider-expected-ns` now does the fallback.

bbatsov:
Looks good to me. :+1:

expez:
Great work @aiba!  :+1:

benedekfazekas:
I bumped the versions for the next dev cycle yesterday. No blocker from my side to merge this.

PR #316, opened 2016-03-26
Remove the warning about missing nREPL ops

benedekfazekas:
follow up CIDER in doing so. Perhaps causes more confusion than
good. Version check is still present.


bbatsov:
The code changes look good to me. In the commit message:

- follow -> Follow
- Perhaps causes -> Probably it causes / Seems it causes
- The final sentence should end with a `.`

expez:
:+1:

bbatsov:
I'll do the honours then. :-)

PR #318, opened 2016-03-27
Enable lexical scoping

bbatsov:
I didn't commit this directly on the slim chance that you might have done this intentionally.

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -1,4 +1,4 @@
-;;; clj-refactor.el --- A collection of clojure refactoring functions
+;;; clj-refactor.el --- A collection of clojure refactoring functions -*- lexical-binding: t -*-
 
 ;; Copyright © 2012-2016 Magnar Sveen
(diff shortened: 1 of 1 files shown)


magnars:
Why enable lexical binding if we're not using it for anything?

bbatsov:
> Why enable lexical binding if we're not using it for anything?

I didn't expect this as I thought the advantages were apparent. :-) E.g. - binding errors, unused vars become obvious, and a certain class of errors (accidental shadowings) become impossible. Pretty much all modern projects use lexical bindings.

Malabarba:
It's also the least surprising option. Lexical scoping is what Clojure does, so it should be more intuitive to the developers here. :-)

magnars:
All good reasons. Thanks!
søn. 27. mar. 2016 kl. 21.19 skrev Artur Malabarba <notifications@github.com

benedekfazekas:
no objection here.

PR #322, opened 2016-04-07
New config setting for libspecs to never prune

expez:
Frontend change matching https://github.com/clojure-emacs/refactor-nrepl/pull/157

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -221,6 +221,15 @@ won't run if there is a broken namespace in the project."
   :group 'cljr
   :type 'boolean)
 
+(defcustom cljr-libspec-whitelist
+  '("^cljsns" "^slingshot.test" "^monger.joda-time" "^monger.json")
+  "List of regexes to match against libspec names which shouldn't be pruned.
+
+This is useful when `clean-ns' should leave a libspec alone even
+if it appears to be unused."
+  :group 'cljr
+  :type '(repeat string))
+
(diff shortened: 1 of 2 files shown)


clj-refactor.el
```diff
@@ -221,6 +221,15 @@ won't run if there is a broken namespace in the project."
   :group 'cljr
   :type 'boolean)
 
+(defcustom cljr-ignore-libspecs
```
expez: I'm quite unhappy with this name.  What do you think @benedekfazekas?

benedekfazekas: thinking along replacing `libspecs` with `ns` but keeping `ignore` but don't have a really good one. perhaps `cljr-ns-to-ignore-patterns`?!

expez: an ns is something else, a libspec may or may not contain a reference to a namespace.

The problem I'm having with the current name ist hat we're not really ignoring them we're just not pruning the occurrences :/

benedekfazekas: hm.. I had not have a proper look but we might want to use this list to ignore stuff for other features as well perhaps?! if so this name is fine in terms of being generic enough

expez: > we might want to use this list to ignore stuff for other features as well perhaps

I don't like complecting things like that :P

expez: @benedekfazekas how about `cljr-libspec-whitelist`?

benedekfazekas:
ok with the name 👍

PR #324, opened 2016-04-07
[Fix #302] Make `ad` work with prismatic/schema

### LINKED ISSUE #302: make `ad` work with schema
It refuses to do anything with `s/defn`

expez:


benedekfazekas:
it seems to me that you submitted this commit in #325 too... I guess that was not the intention?! (see my comments on the code there)

expez:
Actually was intended, as they touched the same code and I tried to reduce the amount of time spent dealing with merge conflicts.

benedekfazekas:
I am a bit confused then. Where do you want to track this commit? in this PR or in #325?

expez:
Here, I'll rebase etc if you want changes to this commit

benedekfazekas:
strangely I can't find that comment of mine. I was only pointing out that perhaps worth mentioning in the changelog that this understands all prefixed `def` like things from now on (prismatic is just an example)

benedekfazekas:
nice :+1:

PR #325, opened 2016-04-07
Give `ad` a prefix variant

expez:
The new prefix variant can be used to declare the thing at point,
instead of the surrounding def.

This is useful for the times when you've just written, or am about to
write, the name of a symbol occuring later in the file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Up next
 
+- [#301] (https://github.com/clojure-emacs/clj-refactor.el/issues/301) `ad`  has gained a prefix to declare the symbol under the cursor.
 - [#312](https://github.com/clojure-emacs/clj-refactor.el/issues/312) Allow `sut` alias to be customized.
 - [#305](https://github.com/clojure-emacs/clj-refactor.el/issues/305) Don't call lookup-alias for non namespaced keywords at all when slash is typed. However trigger lookup alias with the leading :: stripped off the prefix if the keyword is namespaced.
 
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -1430,23 +1430,28 @@ optionally including those that are declared private."
     (re-search-backward def (cljr--point-after 'paredit-backward) :no-error)))
 
 (defun cljr--add-declaration (def)
-  (cljr--goto-declare)
-  (backward-char)
-  (insert " " def)
-  (cljr--post-command-message "Added declaration for %s" def))
+  (when (cljr--already-declared-p def)
+    (user-error "%s is already declared" def))
+  (save-excursion
+    (cljr--goto-declare)
+    (backward-char)
+    (insert " " def)
+    (cljr--post-command-message "Added declaration for %s" def)))
 
 ;;;###autoload
-(defun cljr-add-declaration ()
+(defun cljr-add-declaration (for-thing-at-point-p)
   "Add a declare for the current def near the top of the buffer.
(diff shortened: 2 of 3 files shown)


clj-refactor.el
```diff
@@ -1430,23 +1430,28 @@ optionally including those that are declared private."
...
+    (cljr--post-command-message "Added declaration for %s" def)))
 
 ;;;###autoload
-(defun cljr-add-declaration ()
+(defun cljr-add-declaration (for-thing-at-point-p)
   "Add a declare for the current def near the top of the buffer.
 
+With a prefix add a declaration for the symbol under the cursor instead.
+
 See: https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-add-declaration"
```
benedekfazekas: would be nice to have a gif for the prefix version as well on the wiki

clj-refactor.el
```diff
@@ -1430,23 +1430,28 @@ optionally including those that are declared private."
...
-            (user-error "%s is already declared" def)
-          (cljr--add-declaration def))
-      (user-error "Not inside a def form."))))
+  (interactive "P")
+  (if for-thing-at-point-p
+      (cljr--add-declaration (cider-symbol-at-point))
+    (save-excursion
+      (-if-let (def (cljr--name-of-current-def))
+          (cljr--add-declaration def)
+        (user-error "Not inside a def form.")))))
```
benedekfazekas: wonder if it made sense to default to the `thing-at-point` mode when we are not inside a def instead of giving an error. like the situation in the test. I have no particular usecase in mind tho just a thought.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 
 ### Changes
 
+- [#302] (https://github.com/clojure-emacs/clj-refactor.el/issues/302) `ad` now understands defs using prismatic/schema.
```
benedekfazekas: possibly worth mentioning that it understands all def like macros/things now from libraries referred with an alias.

clj-refactor.el
```diff
@@ -1415,7 +1415,7 @@ optionally including those that are declared private."
 (defun cljr--name-of-current-def ()
   (cljr--goto-toplevel)
   (ignore-errors (forward-char))
-  (when (looking-at "def")
+  (when (looking-at "\\(.+/\\)?def")
```
benedekfazekas: this is not an easy diff to read. generally speaking better to have indentation fix commits separately in the future. not a biggie tho :)

clj-refactor.el
expez: I never write any code outside of some kind of `def` so that code path would never trigger for me.

benedekfazekas: I try to rephrase: if we allow this to run effectively outside a `def` with a prefix why being stricter without a prefix?

expez: As I said, I'm always in the context of a def so it doesn't matter to me. 

If you want to change this to fall back to the other action that's fine, but I don't want to spend time writing code and tests for that now.

benedekfazekas: no worries. I might do this as a separate PR at some point.

benedekfazekas: see my proposed change with tests on this branch `declare-symbol-at-point-outside-def`

expez: Looks good to me

benedekfazekas: coolio, feel free to pull from that branch when you get there

PR #326, opened 2016-04-07
[Fix #299] `ml` moves point

### LINKED ISSUE #299: Cursor moves with `ml`
<kbd>ml</kbd> moves the cursor to the let binding.  When using this refactoring I never actually want to add more bindings so this is getting annoying.

expez:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ### Bugs fixed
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 
 ### Bugs fixed
 
+[#299](https://github.com/clojure-emacs/clj-refactor.el/issues/299) `ml` moves cursor
```
Malabarba: ` - ` missing

PR #328, opened 2016-04-12
Enhance `ad` to declare thing at point outside def

expez:
If we are outside a def we default to declare thing at point as
discussed on #325. The prefix is still needed to declare thing at point
if inside a def.


benedekfazekas:
thanks :+1:

PR #333, opened 2016-05-09
Remove threading related features

benedekfazekas:
because they are moved to Clojure mode. Add minimum clojure-mode version,
keep old memnonics for threading macro related features but point them to
clojure-mode functions.

- [x] You've added tests (if possible) to cover your change(s) (I removed tests this case)


clj-refactor.el
```diff
@@ -8,7 +8,7 @@
 ;;         Benedek Fazekas <benedek.fazekas@gmail.com>
 ;; Version: 2.3.0-SNAPSHOT
 ;; Keywords: convenience, clojure, cider
-;; Package-Requires: ((emacs "24.4") (s "1.8.0") (dash "2.4.0") (yasnippet "0.6.1") (paredit "24") (multiple-cursors "1.2.2") (cider "0.11.0") (edn "1.1.2") (inflections "2.3") (hydra "0.13.2"))
+;; Package-Requires: ((emacs "24.4") (s "1.8.0") (dash "2.4.0") (yasnippet "0.6.1") (paredit "24") (multiple-cursors "1.2.2") (clojure-mode "20160507") (cider "0.11.0") (edn "1.1.2") (inflections "2.3") (hydra "0.13.2"))
```
bbatsov: Guess you can say `clojure-mode` 5.4 here. I can cut a new release if you plan to do a new cljr release soon.

benedekfazekas: that would block us merging this until clojure-mode is released. on the other hand I don't want this to trigger clojure-mode and cljr releases.

So I suppose let's do most of the elisp stuff (as discussed on gitter) and then cut a release of both clojure-mode and cljr.

Malabarba: I'm fine with 20160507 as long as someone remembers to change it to 5.4 before the next stable release of clj-refactor.

Malabarba:
Please use `define-obsolete-variable-alias` for the removed defcustom and `define-obsolete-function-alias` for the removed commands.

expez:
This is the way to go, but I see no rush.  I use these functions every day, and so do probably most of our users.  I'll be changing my bindings manually to help with testing, but I'd like a slightly longer period of testing before we make this change on behalf of everyone else.

benedekfazekas:
just to clarify. I am happy to add these @Malabarba but without readding the defcustom and the functions. If that is what you meant then all 👍.

Malabarba:
>  but without readding the defcustom and the functions.

I don't understand what you mean by "without reading". 
I'm just saying "add aliases pointing to the new commands".

benedekfazekas:
>  I'll be changing my bindings manually to help with testing, but I'd like a slightly longer period of testing before we make this change on behalf of everyone else.

@expez not sure I follow. I intend to merge this as soon as possible as all comments, concerns are resolved I think and I would not like this just hanging around.

Pls note that the usual cljr key combos should just work (only they invoke the reimplemented defuns in Clojure mode).

expez:
@benedekfazekas My point was that that the functions in cljr have been hardened through years of use.  I don't want to inflict a pointless re-implementation on our users until we've had some time to flush out any bugs.

benedekfazekas:
> I don't want to inflict a pointless re-implementation on our users

I could write a really long answer to this but fuck that. Since I am off line next week so I can't fix bugs/regressions if this gets merged on the WE let's agree that I merge it on 23th May if no regressions found in Clojure mode related to it.

expez:
> let's agree that I merge it on 23th May if no regressions found in Clojure mode related to it.

Sounds good.  That probably leaves enough time for testing.

bbatsov:
The new clojure-mode 5.4 is out, btw.

PR #336, opened 2016-06-11
Remove cycle privacy, cycle if, cycle collection

benedekfazekas:
because they are moved to Clojure mode. Add minimum clojure-mode
version (20160604), keep old mnemonics for the first two features and
point them to the clojure-mode functions. Also remove cycle collection
entirely as it got reworked into convert into collection type in Clojure mode.

As the related change got merged into clojure mode a week ago deadline for this is next Sunday.


Malabarba:
> Also remove cycle collection
entirely as it got reworked into convert into collection type in Clojure mode.

It's up to you, but it might be good to keep this command. I just know there someone who's gonna complain they liked the old one better.

expez:
LGTM :+1:

benedekfazekas:
@Malabarba I was considering keeping the cycle collection types here by rewriting the function so it calls the right fns in Clojure mode. Still happy to do it but thinking that the new solution is actually better in the sense of being simpler than cycling. So why not push users towards it a bit?

Malabarba:
> I was considering keeping the cycle collection types here by rewriting the function so it calls the right fns in Clojure mode. 

Nah, I think it's not worth the trouble.

PR #338, opened 2016-06-21
Fix that cljr--clojure-filename-p didn't use its arg

robert-stuttaford:

- [ ] The commits are consistent with our [contribution guidelines](CONTRIBUTING.md)

Thanks!


expez:
Thanks! :+1:

PR #347, opened 2016-10-05
[Fix #341] Avoid circular dep when using mv

### LINKED ISSUE #341: cljr-move-form handles refers badly
## Expected behavior

When moving a form that has functions aliased from the target namespace, I expect the aliases to be removed from inside the form.

## Actual behavior

Instead, the target namespace's ns form is altered to require itself, creating an illegal circular dependency.

Start with the following two namespaces:

```
(ns origin
  (:require [target :as t]))

(defn doit [x]
  (t/really-do-it x))
```

```
(ns target)

(defn really-do-it [x] (println x))
```

Put the point at the opening paren of origin/doit. Execute C-c C-a mf "target.clj".

The resulting namespace will be:

```
(ns target
  (:require  [target :as t]))

(defn really-do-it [x] (println x))

(defn doit [x]
  (t/really-do-it x))
```

Which will not compile.

clj-refactor 2.2.0, refactor-nrepl 2.2.0

### CIDER version information

CIDER 0.12.0 (Seattle)

### Leiningen or Boot version

```
#http://boot-clj.com
#Sun Jul 31 13:26:32 CDT 2016
BOOT_CLOJURE_NAME=org.clojure/clojure
BOOT_CLOJURE_VERSION=1.9.0-alpha8
BOOT_VERSION=2.6.0
```

### Emacs version

```
GNU Emacs 24.5.1 (x86_64-apple-darwin14.3.0, NS apple-appkit-1347.57) of 2015-05-28 on coil.local
```

### Operating system

OS X 10.11 "El Capitan"

benedekfazekas:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,8 @@
 - [#309](https://github.com/clojure-emacs/clj-refactor.el/issues/309)  `am` creates alias for fully-qualified symbols.
 - [#313](https://github.com/clojure-emacs/clj-refactor.el/issues/313)  teach `pf` about function literals using `%&`.
 - [#320](https://github.com/clojure-emacs/clj-refactor.el/issues/320) `*data-readers*` ignored when searching for macros.
-- [#339](https://github.com/clojure-emacs/clj-refactor.el/issues/339) Teach stop refer to understand multiline refer clauses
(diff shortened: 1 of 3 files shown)


expez:
Yay! \o/

PR #351, opened 2016-10-18
Add space after dashes-to-underscores prompt

raxod502:
This changes the prompt from "Replace with underscores?(yes or no)" to "Replace with underscores? (yes or no)", which looks more professional.

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -1050,7 +1050,7 @@ word test in it and whether the file lives under the test/ directory."
 
 (defun cljr--maybe-replace-dash-in-file-name (file-name)
   (if (and (cljr--dash-in-file-name-p file-name)
-           (yes-or-no-p "The file name contains dashes. Replace with underscores?"))
+           (yes-or-no-p "The file name contains dashes. Replace with underscores? "))
       (concat (file-name-directory file-name)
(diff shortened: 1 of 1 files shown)


expez:
Cool, thanks!  Not sure why that broke our build :/

raxod502:
Any update on this?

expez:
I've been pretty swamped at work so I haven't been able to look into the failing build.  I also didn't want to turn our build badge red to improve the formatting slightly.

@benedekfazekas Do you have time to investigate?

benedekfazekas:
will try to check this week at some point

benedekfazekas:
the failure seems to be in cider build which fails cljr build too. I also had a look on the cider build but got pretty confused by the failure there...

i guess no point to rerun this until cider build gets fixed.

benedekfazekas:
it seems reason is that support for emacs 24.4 and 24.5 was accidentally dropped in cider (intention was to drop support for 24.3)

benedekfazekas:
@raxod502 please sync with master (cider got fixed, but also had our own problems which got fixed too). so after sync the build should pass for this PR too

expez:
Thanks for this @raxod502 :+1: 

And @benedekfazekas for digging into the build failure!

PR #354, opened 2016-11-25
use cl-assert instead of assert

bendlas:
clj-refactor.el assumes the 'cl package to be required, as it uses its `assert` macro.
That assumption turns out to be false with my recent version of cider and emacs.

- `M-x emacs-version`
  `GNU Emacs 25.1.1 (x86_64-unknown-linux-gnu, GTK+ Version 3.22.1) of 2016-11-22`
- `M-x cider-version`
  `CIDER 0.14.0 (Berlin)`
- From a clojure file `M-x cider-jack-in`
  ```
  Starting nREPL server via /run/current-system/sw/bin/lein update-in :dependencies conj [org.clojure/tools.nrepl\ \"0.2.12\"\ \:exclusions\ \[org.clojure/clojure\]\] -- update-in :plugins conj \[refactor-nrepl\ \"2.2.0\"\] -- update-in :plugins conj \[cider/cider-nrepl\ \"0.14.0\"\] -- repl :headless...
  nREPL server started on 39893
  [nREPL] Establishing direct connection to localhost:39893 ...
  [nREPL] Direct connection established
  error in process filter: cljr--create-msg: Symbol’s function definition is void: assert
  error in process filter: Symbol’s function definition is void: assert
```

### DIFF
--- a/Cask
+++ b/Cask
@@ -8,4 +8,5 @@
   (depends-on "espuds")
   (depends-on "cider")
   (depends-on "edn")
-  (depends-on "clojure-mode"))
+  (depends-on "clojure-mode")
+  (depends-on "cl-lib"))
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -40,6 +40,7 @@
 (require 'sgml-mode)
 (require 'inflections)
 (require 'hydra)
+(require 'cl-lib)
 
 (defcustom cljr-add-ns-to-blank-clj-files t
   "If t, automatically add a ns form to new .clj files."
@@ -287,7 +288,7 @@ Otherwise open the file and do the changes non-interactively."
        (not (bound-and-true-p evil-mode))))
 
 (defun cljr--fix-special-modifier-combinations (key)
-  (case key
+  (cl-case key
     ("C-s-i" "s-TAB")
     ("C-s-m" "s-RET")
     (otherwise key)))
@@ -714,7 +715,7 @@ at the opening parentheses of an anonymous function."
   the kv pairs KVS.
 
 All config settings are included in the created msg."
-  (assert (cljr--evenp (length kvs)) nil "Can't create msg to send to the middleware.\
+  (cl-assert (cljr--evenp (length kvs)) nil "Can't create msg to send to the middleware.\
   Received an uneven number of kv pairs: %s " kvs)
   (apply #'list "op" op
          "prefix-rewriting"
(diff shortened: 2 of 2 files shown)


bendlas:
Wait a minute ... this is probably a bug in 'cl-lib in my emacs version. Code is not supposed to require 'cl directly, but 'cl-lib, which should autoload everything, but doesn't load assert for me ... investigating ...

bendlas:
**EDIT** updated title from `require 'cl package for assert macro`

Not a bug in 'cl-lib or emacs. 'cl-lib doesn't export `assert` any more.
Pushed an updated commit to utilize 'cl-lib instead of deprecated 'cl

benedekfazekas:
thanks.

build fails due to cider having an issue with emacs 24.x. let's wait until cider gets fixed and see if the build is fine.

benedekfazekas:
still fails on 24.4 and 24.5 unfortunately

bendlas:
Too bad. The error seems unrelated to this PR though, as the other PR fails with the exact same message.

benedekfazekas:
I think it fails on a `case` that is coming from `cl-lib` too, isn't it?

bendlas:
looks to me that the culprit is something else: https://travis-ci.org/clojure-emacs/clj-refactor.el/jobs/178758356#L708

benedekfazekas:
maybe I am wrong but that is inside a `case` that is the reason it fails...

bendlas:
ok, now the basic build errors are gone but there are a bunch of failing test cases

benedekfazekas:
hm... perhaps api changes in those functions that got updated?!

bendlas:
Weird thing is: when I eval the updated .el and test some of the failing cases, they work fine.
Could that have something to do with .elc compilation? What versions of cider, ... does travis use?

benedekfazekas:
uses cask to retrieve latest HEAD. it should be also there in the build log. thing is: it does not really matter what works in your local emacs installation. have a look in the build script in root.

benedekfazekas:
thanks btw for taking the pain here. appreciated

benedekfazekas:
best guess that this is the problem: https://github.com/eschulte/jump.el/pull/11

at this point I see two options: either kill off the `inflections` calls from the code or wait for the above to me merged...

benedekfazekas:
any chance @purcell you can have a look at https://github.com/eschulte/jump.el/pull/11

purcell:
@benedekfazekas 👍

benedekfazekas:
thanks @purcell 

@bendlas I think this is good now, there are still failures but I think those are unrelated to the `cl-lib` vs `cl` issue. can you please squash the above commits tho?

benedekfazekas:
ta @bendlas

PR #355, opened 2016-12-23
Remove `let` features related code where possible

benedekfazekas:
Remove all `let` related code where possible. Multiple cursors related
code remains here. If multiple cursors is not used defer to clojure
mode implementation of the feature. Additionally use functions migrated
to clojure mode whereever possible.

Compatible with clojure mode 5.6.1

Remove some tests duplicated in clojure mode. Only those tests remain
which use multiple cursors. The ones without multiple cursors are
removed as they are test the same code as the ones in clojure mode.

Introduce let accepts prefix arg so it can pass it on to clojure-mode
implementation.


expez:
:+1:

PR #357, opened 2017-01-13
Cygwin path fix.

tsulej:
See error description [here](https://github.com/clojure-emacs/refactor-nrepl/issues/183).

Context: Emacs under Cygwin generates unix-like paths which are not accepted by nREPL. This error is visible when running `replace symbol`.
Cider provides two functions: `cider-to-nrepl-filename-function` and `cider-from-nrepl-filename-function` to make proper conversions. These functions make conversions only when you run Emacs under Cygwin.

Following fix wraps `(buffer-file-name)` when creating request.
To process middleware response I introduced `cljr--get-valid-filename` function which extracts `:file` from the hash and converts path to back to unix-like.

I tested only `replace symbol` case.


CHANGELOG.md
```diff
@@ -25,6 +25,7 @@
 - [#320](https://github.com/clojure-emacs/clj-refactor.el/issues/320) `*data-readers*` ignored when searching for macros.
 - [#339](https://github.com/clojure-emacs/clj-refactor.el/issues/339) Teach stop refer to understand multiline refer clauses.
 - [#341](https://github.com/clojure-emacs/clj-refactor.el/issues/341) Avoid creating circular dependencies with move form when source namespace refers to the target namespace with a require.
+- Cygwin path fix. Conversion between Unix-like path generated by Emacs and Windows-like path required by nREPL.
```
benedekfazekas: please reference the `refactor-nrepl` issue you created (see examples of this below in the changelog). also perhaps mention the fact that a cider solution is reused.

clj-refactor.el
```diff
@@ -2401,8 +2401,12 @@ See: https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-promote-function
 root."
   (s-chop-prefix (cljr--project-dir) path))
 
+(defun cljr--get-valid-filename (hash)
+    "Get :file value from the hash table and convert path if necessary. Cygwin fix."
```
benedekfazekas: I don't think "Cygwin fix". is needed here.

clj-refactor.el
```diff
@@ -2350,7 +2350,7 @@ See: https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-promote-function
   (= (point) (cljr--point-after 'end-of-buffer)))
 
 (defun cljr--find-symbol-sync (symbol ns)
-  (let* ((filename (buffer-file-name))
+  (let* ((filename (funcall cider-to-nrepl-filename-function (buffer-file-name)))
```
expez: You can delete all occurrences of `funcall`.  If you know the name of the function you can call it like this:

`(cider-to-nrepl-filename-function arg-goes-here)`

expez:
Have you tested all this manually?  We don't have any real integration tests testing the entire stack, unfortunately.

clj-refactor.el
tsulej: Unfortunately not. These two functions are defined as `defvar` and calling directly causes the error.

CHANGELOG.md
tsulej: Sure!

clj-refactor.el
tsulej: ok

tsulej:
Yes, I've tested this manually. Only one case `replace symbol`. If you can point me to other functions which require nrepl interaction with filename inside I'll test them either.
Fortunately it should be safe fix, outside cygwin `cider-to-nrepl-filename-function` and `cider-from-nrepl-filename-function` is `identity`. Updating code after review.

benedekfazekas:
thanks for the quick and nice fix @tsulej

PR #360, opened 2017-01-22
Fix cljr--normalize-symbol-name for `#'` prefix

areina:
Hi! There are no tests for this function, so I tested it directly through `ielm`:

```elisp
;; Before
ELISP> (cljr--normalize-symbol-name "#'sym")
""
;; After
ELISP> (cljr--normalize-symbol-name "#'sym")
"sym"
```

- [X] You've updated the changelog (if adding/changing user-visible functionality)

Thanks!


expez:
Thanks! :+1:

Issue #293, opened 2016-02-11
Drop the s.el dependency


bbatsov:
I see you're depending on Emacs 24.4. I'd suggest dropping s.el and simply using `subr-x.el` introduced there. The fewer deps, the better.

expez:
Pull request welcome :)

expez:
Awesome, @dotemacs :+1:

benedekfazekas:
looks cool!

will try to find time to review in detail but so far looks really good. also not planning to merge this before releasing `2.3.0`

benedekfazekas:
what is the problem with using `thread-last` here? https://github.com/clojure-emacs/clj-refactor.el/compare/master...dotemacs:no-more-s#diff-ad8315941dfda54547f9762bd5a52d04L305

dotemacs:
Hey @benedekfazekas 

I knew that this question would come up :)

So I put the comment about it in the commit, pasting it here so that you don't have to go looking for it

> All the `thread-last` got unwound due to the re-write as s.el's `s-join`
expected separator first then the strings. Whereas subr-x.el's
`string-join` expects strings and then the separator.

https://github.com/dotemacs/clj-refactor.el/commit/da35e4ba51e0bda6139f47aec1290eab786f8d5d

So it was a case of should I keep `thread-last` just to keep it? Or should I write a function that uses `string-join` internally but takes arguments as `s-join` and reorders them for `string-join`?
I figured that just unwinding `thread-last` was the easiest option.
But if you have a prefered way of doing this, then please do share :)


Also, there are 2-3 functions more from s.el that I still need to re-write, that's why I haven't created the pull request just yet. I only put a comment here that I'm working on it, so that somebody else doesn't pick it up... 

Should I create a pull request now, if we're to have more discussion about the code?

As for when you'll merge this, no rush on my side.

benedekfazekas:
ha, makes sense. sorry missed the commit comment.

PR: yes please. would be easier to comment stuff, etc.

PR #372, opened 2017-04-27
Removal of s.el as a dependency

dotemacs:
This PR based on the this issue https://github.com/clojure-emacs/clj-refactor.el/issues/293 "Removal of s.el as a dependency".

The dependency on s.el has now been removed.

Please provide your feedback about the work I've done so far.

---- 

**Regarding the PR etiquette below, I've crossed out the sections that I don't think are applicable.**

- [x] The commits are consistent with our [contribution guidelines](https://github.com/clojure-emacs/clj-refactor.el/blob/master/.github/CONTRIBUTING.md)
~~- [ ] You've updated the changelog (if adding/changing user-visible functionality)~~
~~- [ ] You've updated the readme (if adding/changing user-visible functionality)~~

Thanks!

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -761,14 +761,17 @@ A new record is created to define this constructor."
 ;; ------ file -----------
 
 (defun cljr--project-dir ()
-  (or (ignore-errors
-        (file-truename
-         (locate-dominating-file default-directory "project.clj")))
-      (ignore-errors
-        (file-truename
-         (locate-dominating-file default-directory "build.boot")))
-      (ignore-errors (file-truename
-                      (locate-dominating-file default-directory "pom.xml")))))
+  (let ((project-dir (or (ignore-errors
+			   (file-truename
+			    (locate-dominating-file default-directory "project.clj")))
+			 (ignore-errors
+			   (file-truename
+			    (locate-dominating-file default-directory "build.boot")))
+			 (ignore-errors (file-truename
+					 (locate-dominating-file default-directory "pom.xml"))))))
+    (if (null project-dir)
+	""
+      project-dir)))
 
 (defun cljr--project-file ()
   (let ((project-dir (cljr--project-dir)))
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -41,6 +41,7 @@
 (require 'sgml-mode)
 (require 'inflections)
 (require 'hydra)
+(require 'subword)
 
 (defcustom cljr-add-ns-to-blank-clj-files t
   "If t, automatically add a ns form to new .clj files."
@@ -3365,6 +3366,17 @@ See: https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-create-fn-from-e
   (when (string-match "[^/]+$" s)
     (substring s (car (match-data)) (car (cdr (match-data))))))
 
+(defun cljr--dashed-words (s)
+  "Take the string S and replace all the word separators with '-'
+and make the whole string lower-cased."
+  (with-temp-buffer
+    (insert s)
+    (goto-char (point-min))
+    (while (not (eobp))
+      (subword-forward)
+      (insert " "))
+    (mapconcat 'identity (split-string (downcase (buffer-string))) "-")))
+
 (defun cljr--guess-param-name (form)
   (let* ((prepped-form (cljr--strip-off-semantic-noops
                         (cljr--unwind-s form)))
@@ -3375,11 +3387,11 @@ See: https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-create-fn-from-e
      ((cljr--keyword-lookup-p prepped-form)
       (cljr--strip-keyword-ns (match-string 1 prepped-form)))
      ((and fn-call (string-suffix-p "." fn-call))
-      (s-dashed-words (car (last (split-string fn-call "\\." t)))))
+      (cljr--dashed-words (car (last (split-string fn-call "\\." t)))))
      ((and fn-call (string-prefix-p "create-" fn-call))
       (string-remove-prefix "create-" fn-call))
      ((and fn-call (string-prefix-p ".get" fn-call))
-      (s-dashed-words (string-remove-prefix ".get" fn-call)))
+      (cljr--dashed-words (string-remove-prefix ".get" fn-call)))
      ((string= "get-in" fn-call)
       (cljr--find-param-name-from-get-in prepped-form))
      ((string= "get" fn-call)
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -1323,7 +1323,7 @@ See: https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-move-form"
            (refer-present-p (cljr--search-forward-within-sexp ":refer" :save-excursion))
            (refer-all-p (cljr--search-forward-within-sexp ":refer :all" :save-excursion))
            (require-present-p (cljr--search-forward-within-sexp
-                               (s-concat ":require [" ns)
+                               (concat ":require [" ns)
                                :save-excursion)))
       (if ns-present-p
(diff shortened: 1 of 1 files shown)


clj-refactor.el
```diff
@@ -3365,6 +3366,17 @@ See: https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-create-fn-from-e
...
+(defun cljr--dashed-words (s)
+  "Take the string S and replace all the word separators with '-'
+and make the whole string lower-cased."
+  (with-temp-buffer
+    (insert s)
+    (beginning-of-buffer)
+    (while (not (eobp))
+      (subword-forward)
+      (insert " "))
+    (mapconcat 'identity (split-string (downcase (buffer-string))) "-")))
```
dotemacs: What do you think of this?

I'm a bit undecided. 

In one way it only uses functions that come as part of Emacs and relies on `subword` mode. But a bit unsure about it... suggestions welcome.

expez: Looks OK to me.  If it works as it should I don't see a huge point in bikeshedding implementation details we can change at will :)

dotemacs: > Looks OK to me.

Cool, glad that you approve :)

dotemacs:
The only functions left so far from **s.el** are `s-slice-at` & `s-with`. 
They are both used only once.

I'll try to tackle those tonight/tomorrow...

dotemacs:
Only `s-slice-at` is left...

dotemacs:
I've removed all the functions from `s.el` and used stock ones from Emacs.

I've updated the PR description with some comments, please read that first as there are a few outstanding issues that I require your feedback on.

Then please review the code.

Thank you :)

benedekfazekas:
to  address your comments above in bold:

contributing guidelines here: https://github.com/clojure-emacs/clj-refactor.el/blob/master/.github/CONTRIBUTING.md perhaps the link is wrong in the PR template?!

the README also has the (correct) link

tests: this is internal refactoring i guess we need to rely on our existing test set to catch regressions... don't think you need to add new unit like tests for this.

clj-refactor.el
```diff
@@ -573,7 +573,7 @@ list of (fn args) to pass to `apply''"
 
 (defun cljr--whitespacep (s)
   "True if S contains only whitespace."
-  (s-blank? (string-trim s)))
```
benedekfazekas: essential difference between `s-blank?` and `string-blank-p` that former is null safe latter is not. I don't think that causes any problem in this commit but you might want to dblcheck that..

clj-refactor.el
```diff
@@ -761,14 +761,17 @@ A new record is created to define this constructor."
...
-                      (locate-dominating-file default-directory "pom.xml")))))
+  (let ((project-dir (or (ignore-errors
+			   (file-truename
+			    (locate-dominating-file default-directory "project.clj")))
+			 (ignore-errors
+			   (file-truename
+			    (locate-dominating-file default-directory "build.boot")))
+			 (ignore-errors (file-truename
+					 (locate-dominating-file default-directory "pom.xml"))))))
+    (if (null project-dir)
```
benedekfazekas: perhaps rather `(or project-dir "")`?

or in fact you can get rid of the `let` you introduced and just stick an empty string at the end of the original `or` expression as default?

clj-refactor.el
```diff
@@ -2689,9 +2689,9 @@ str/split => split"
   (let ((name (cljr--normalize-symbol-name symbol)))
     (cond
      ((string-match-p "\\w+\\.\\w+" name)
-      (thread-last name (s-split "\\.") last car cljr--symbol-suffix))
+      (thread-last name (split-string "\\.") last car cljr--symbol-suffix))
```
benedekfazekas: this should be `thread-first` as well, or?

clj-refactor.el
```diff
@@ -2247,15 +2247,15 @@ See: https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-update-project-d
               (insert name)
             (mc/create-fake-cursor-at-point)))
         (re-search-forward "\\[")
-        (when (s-present? locals)
+	(when (not (or (null locals) (string= "" locals)))
```
benedekfazekas: i would possibly rather go for
`string-empty-p` instead of `string= ""` for readability and perhaps factor this pred out in its own function? something like `cljr--string-present-p`?

benedekfazekas:
made couple of comments. hope they make sense. only one of those is a potential bug the rest is more like style etc things.

clj-refactor.el
dotemacs: Good point.

Changed it as per your comment.

I was more on the 'drive' to complete the whole task, so I've missed the obvious stuff like this. Thanks

clj-refactor.el
dotemacs: Created `cljr--string-present-p` as per your comment...

clj-refactor.el
dotemacs: Good catch. Fixed.

Shame that the tests didn't catch this...

clj-refactor.el
dotemacs: Good point, added the null safety with 36a886b

clj-refactor.el
benedekfazekas: tests: indeed

dotemacs:
@benedekfazekas addressed all your comments.

Please review again and let me know what you think.

Thanks :)

clj-refactor.el
```diff
@@ -769,9 +769,7 @@ A new record is created to define this constructor."
 			    (locate-dominating-file default-directory "build.boot")))
 			 (ignore-errors (file-truename
 					 (locate-dominating-file default-directory "pom.xml"))))))
-    (if (null project-dir)
-	""
-      project-dir)))
+    (or project-dir "")))
```
benedekfazekas: do you need the `let` at all tho? you can just
```
(or (ignore-errors...)
     (ignore-errors...)
     (ignore-errors...)
     "")
```

or am i missing something?

benedekfazekas:
made one more comment. otherwise LGTM. great work!

benedekfazekas:
one more thing: `cljr--string-present-p` could be used at more places i am guessing. for example line 2264 and 3207

clj-refactor.el
dotemacs: I'd happily add the test for this, but seeing as `cljr--symbol-suffix` is a non-interactive function and most of the tests are feature tests, should I have a step like this:

```feature
When I press "M-: (cljr--symbol-suffix \"java.util.Date\")"
Then I should see message "Date"
```

What are your thoughts?

benedekfazekas: we decided to go down the behavioural testing route not the unit testing one. which is essence was a good choice i think. so i suppose just leave it ;)

dotemacs:
Reused `cljr--string-present-p` where suitable... 1641730
Removed let c02c15f & reduced repetition 4c0b23b

See what you think of it now. 

Thanks for the feedback

clj-refactor.el
```diff
@@ -760,16 +760,18 @@ A new record is created to define this constructor."
...
-      (ignore-errors
-	(file-truename
-	 (locate-dominating-file default-directory "build.boot")))
-      (ignore-errors (file-truename
-		      (locate-dominating-file default-directory "pom.xml")))
-      ""))
+  (or
+   (thread-last  '("project.clj" "build.boot" "pom.xml")
+     (mapcar 'cljr--locate-project-file)
+     (delete '"")
```
benedekfazekas: really like this approach. but i think this line is wrong. `cljr--locate-project-file` returns nil not empty string if file is not present. should be `(delete nil)` instead.

dotemacs: Thanks, pushed it as part of c33fc5f

benedekfazekas:
great work @dotemacs ta!

PR #392, opened 2017-10-15
[Fix #391] Add option for disabling warnings

### LINKED ISSUE #391: Can clj-refactor.el degrade cleanly outside a project context?
By default, when I enable `clj-refactor-mode` in all `clojure-mode` buffers as recommended, and I start a `M-x cider-jack-in`, I receive the following in the `*cider-repl*` buffer:

```
WARNING: clj-refactor and refactor-nrepl are out of sync.
Their versions are 2.4.0-SNAPSHOT and n/a, respectively.
You can mute this warning by changing cljr-suppress-middleware-warnings.
```

And I receive the following in the `*nrepl-server*` buffer:

```
Warning: refactor-nrepl needs to run in the context of a project.
Warning: refactor-nrepl middleware won't be activated.
Warning: refactor-nrepl needs to run in the context of a project.
Warning: refactor-nrepl middleware won't be activated.
Warning: refactor-nrepl needs to run in the context of a project.
Warning: refactor-nrepl middleware won't be activated.
Warning: refactor-nrepl needs to run in the context of a project.
Warning: refactor-nrepl middleware won't be activated.
```

However, none of this is useful. Of course since I ran `M-x cider-jack-in` outside the context of a project, refactor-nrepl won't be available. I would like the option to suppress all warnings related to the situation, without also suppressing warnings from actual erroneous behavior.

The message in `*cider-repl*` suggests to customize `cljr-suppress-middleware-warnings`. However:

* that will suppress legitimate warnings about e.g. the wrong Clojure version being used, or refactor-nrepl can't be loaded when inside a project context
* it doesn't suppress the warnings in `*nrepl-server*`

The correct solution, in my opinion, is to _not inject refactor-nrepl at all_ unless the predicate `(cljr--project-dir)` returns non-nil. I have a [rather elaborate hack](https://github.com/raxod502/radian/blob/ee92ea6cb0473bf7d20c6d381753011312ef4a52/radian-emacs/radian-clojure.el#L339-L401) in my init-file to accomplish exactly this, but I think it'd be a better experience for all users if such behavior were the default.

raxod502:
Add `cljr-suppress-outside-project-warning` defcustom. Set `cider-lein-*` variables using advice-driven let-binding instead of global modification, and only inject nrepl-middleware when inside project context. After REPL has started, if outside project context, display more specific warning (unless the new defcustom is set to non-nil).

    - **No, as existing warnings were not tested**
    - **No, as existing variable `cljr-suppress-middleware-warnings` was not documented there**

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -8,7 +8,7 @@
 ;;         Benedek Fazekas <benedek.fazekas@gmail.com>
 ;; Version: 2.4.0-SNAPSHOT
 ;; Keywords: convenience, clojure, cider
-;; Package-Requires: ((emacs "24.4") (s "1.8.0") (seq "2.19") (yasnippet "0.6.1") (paredit "24") (multiple-cursors "1.2.2") (clojure-mode "5.6.1") (cider "0.15.0") (edn "1.1.2") (inflections "2.3") (hydra "0.13.2"))
+;; Package-Requires: ((emacs "24.4") (s "1.8.0") (seq "2.19") (yasnippet "0.6.1") (paredit "24") (multiple-cursors "1.2.2") (clojure-mode "5.6.1") (cider "0.17.0") (edn "1.1.2") (inflections "2.3") (hydra "0.13.2"))
 
 ;; This program is free software; you can redistribute it and/or
 ;; modify it under the terms of the GNU General Public License
@@ -152,6 +152,13 @@ as can be."
 
 (defcustom cljr-suppress-middleware-warnings nil
   "If t, no middleware warnings are printed to the repl."
+  :group 'cljr
+  :type 'boolean)
+
+(defcustom cljr-suppress-no-project-warning nil
+  "If t, no warning is printed when starting a REPL outside a project.
+By default, a warning is printed in this case since clj-refactor
+will not work as expected in such REPLs."
   :group 'cljr
   :type 'boolean)
 
@@ -785,6 +792,10 @@ A new record is created to define this constructor."
      (delete 'nil)
      car)
    ""))
+
+(defun cljr--inside-project-p ()
+  "Return non-nil if `default-directory' is inside a Clojure project."
+  (not (string-empty-p (cljr--project-dir))))
 
 (defun cljr--project-file ()
   (let ((project-dir (cljr--project-dir)))
@@ -3142,15 +3153,21 @@ if REMOVE-PACKAGE_VERSION is t get rid of the (package: 20150828.1048) suffix."
 
 (defun cljr--check-middleware-version ()
   "Check whether clj-refactor and nrepl-refactor versions are the same"
-  (let ((refactor-nrepl-version (or (cljr--middleware-version)
-                                    "n/a")))
-    (unless (string-equal (downcase refactor-nrepl-version)
-                          (downcase (cljr--version :remove-package-version)))
-      (cider-repl-emit-interactive-stderr
-       (format "WARNING: clj-refactor and refactor-nrepl are out of sync.
+  (if (cljr--inside-project-p)
+      (let ((refactor-nrepl-version (or (cljr--middleware-version)
+                                        "n/a")))
+        (unless (string-equal (downcase refactor-nrepl-version)
+                              (downcase (cljr--version :remove-package-version)))
+          (cider-repl-emit-interactive-stderr
+           (format "WARNING: clj-refactor and refactor-nrepl are out of sync.
 Their versions are %s and %s, respectively.
 You can mute this warning by changing cljr-suppress-middleware-warnings."
-               (cljr--version) refactor-nrepl-version)))))
+                   (cljr--version) refactor-nrepl-version))))
+    (unless cljr-suppress-no-project-warning
+      (cider-repl-emit-interactive-stderr
+       "WARNING: No Clojure project was detected. The
+refactor-nrepl middleware was not enabled. (You can mute this
+warning by customizing `cljr-suppress-no-project-warning'.)"))))
 
 ;;;###autoload
 (defun cljr-version ()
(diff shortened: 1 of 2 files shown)


raxod502:
I don't know why the build is failing, but it does not seem relevant to my change.

clj-refactor.el
```diff
@@ -3992,18 +4000,23 @@ See: https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-change-function-
...
+  (if (and (not (string-empty-p (cljr--project-dir)))
+           cljr-inject-dependencies-at-jack-in)
+      (let ((cider-jack-in-lein-plugins
+             (cons `("refactor-nrepl" ,(cljr--version t))
+                   cider-jack-in-lein-plugins))
+            (cider-jack-in-nrepl-middlewares
+             (cons "refactor-nrepl.middleware/wrap-refactor"
+                   cider-jack-in-nrepl-middlewares)))
+        (apply cider-jack-in args))
+    (apply cider-jack-in args)))
```
benedekfazekas: why do you prefer `cons` over `add-to-list`?

clj-refactor.el
```diff
@@ -3992,18 +4000,23 @@ See: https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-change-function-
...
+(defun cljr--inject-jack-in-dependencies (cider-jack-in &rest args)
   "Inject the REPL dependencies of clj-refactor at `cider-jack-in'.
 If injecting the dependencies is not preferred set `cljr-inject-dependencies-at-jack-in' to nil."
-  (when (and cljr-inject-dependencies-at-jack-in
-             (boundp 'cider-jack-in-lein-plugins)
-             (boundp 'cider-jack-in-nrepl-middlewares))
-    (add-to-list 'cider-jack-in-lein-plugins `("refactor-nrepl" ,(cljr--version t)))
-    (add-to-list 'cider-jack-in-nrepl-middlewares "refactor-nrepl.middleware/wrap-refactor")))
+  (if (and (not (string-empty-p (cljr--project-dir)))
+           cljr-inject-dependencies-at-jack-in)
```
benedekfazekas: keeping the `boundp` checks on the cider vars make sense to make sure that the cider version installed is compatible with this feature... any particular reason you got rid of them?

benedekfazekas:
apart from the comments i am 👍  to merge this. build failure is weird indeed. perhaps something with `cask`?!

clj-refactor.el
raxod502: Because `add-to-list` would modify the list variable globally, unless I `let`-bound the variable to itself (which would be more complicated than just using `cons`).

clj-refactor.el
raxod502: Oops, I was originally thinking we didn't need them since I switched from `add-to-list` to `let` + `cons`, but we still do since I use the original value. Fixing that now.

clj-refactor.el
```diff
@@ -3992,18 +4000,25 @@ See: https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-change-function-
...
+            (cider-jack-in-nrepl-middlewares
+             (cons "refactor-nrepl.middleware/wrap-refactor"
+                   cider-jack-in-nrepl-middlewares)))
+        (apply cider-jack-in args))
+    (apply cider-jack-in args)))
 
 ;;;###autoload
 (eval-after-load 'cider
-  '(cljr--inject-jack-in-dependencies))
+  '(advice-add #'cider-jack-in :around #'cljr--inject-jack-in-dependencies))
```
expez: @bbatsov This seems like a no-no to me.  What are you thoughts on using an advice hack here?

For context, @raxod502 is submitting a patch so a variable will prevent the middleware dep from even being injected when we're outside a project.  The reason being that the middleware will emit warnings that he doesn't care about.

clj-refactor.el
```diff
@@ -155,6 +155,14 @@ as can be."
   :group 'cljr
   :type 'boolean)
 
+(defcustom cljr-suppress-outside-project-warning nil
```
expez: I think `cljr-suppress-no-project-warning` might be a bit more self-explanatory.

clj-refactor.el
```diff
@@ -155,6 +155,14 @@ as can be."
...
+  "If t, no warning is printed when starting a REPL outside a project.
+By default, a warning is printing in this case since clj-refactor
+will not work as expected in such REPLs and therefore the
```
expez: Cut everything after "and therefore", as that last part only confuses things.

clj-refactor.el
raxod502: The reason I used advice here was that advice was already being used elsewhere; see https://github.com/clojure-emacs/clj-refactor.el/blob/f5295df68955c23fffd60718039fd386d13c77f5/clj-refactor.el#L1638.

Note also that mutating a global variable in an autoload is not considered good practice either, so I don't think either the previous code or the new code is ideal.

clj-refactor.el
raxod502: Good idea. Changed.

clj-refactor.el
raxod502: Done.

clj-refactor.el
raxod502: If you really don't want to use advice, I'll be happy to submit a pull request upstream for some additional customizable variables to be added so that doing this doesn't require advice.

clj-refactor.el
benedekfazekas: I see no harm in modifying these variables globally. Am I missing any edge cases here?

raxod502: If the variables were modified globally, then that would cause refactor-nrepl to be injected always, as opposed to just when you are inside a project. Avoiding that behavior is the whole point of this patch.

raxod502:
Any update here?

expez:
> Any update here?

We had another user requesting this recently :)

I'd like to take you up on the offer to submit a PR to CIDER to avoid the use of advice.

You're right that we have some advice for paredit, but if that advice goes stale, and breaks, everything else works fine, whereas if the advice for cider breaks down, due to changes upstream, we're likely to break both CIDER and clj-refactor for everyone.

The changelog entry is also added in the wrong place (in the release notes for an already released version).  That's on us, I guess, for not pushing a an 'Upcoming' section.  Sorry about that.

bbatsov:
Indeed!

raxod502:
I've created a pull request at https://github.com/clojure-emacs/cider/pull/2238. Apologies for the delay in doing so.

raxod502:
I've also updated this pull request to use the `:predicate` feature introduced therein, so advice is no longer required.

clj-refactor.el
```diff
@@ -786,6 +793,10 @@ A new record is created to define this constructor."
      car)
    ""))
 
+(defun cljr--inside-project-p ()
```
expez: Think this should be called `cljr--inject-middleware-p?` to better reflect its usage.   

Also, shouldn't it have an argument list like `(&rest _)` since it's actually expecting 3 arguments, that are ignored?

Edit: I see it's used in two places, so I'd suggest leaving this and creating another function, `cljr--inject-middleware-p` that just has the right argument list and calls out to this one.

raxod502:
Thanks, fixed.

raxod502:
@expez Now that the pull request for CIDER is functional and tested, I fixed a few things here and tested. Appears to be working.

expez:
Great! :+1:

raxod502:
@expez Well, the pull request has been merged upstream. Now we can look at this one :)

expez:
Great work @raxod502!  

Thanks for seeing this through all the way! :+1:

Issue #353, opened 2016-10-27
"Invalid token" during cljr-clean-ns for keywords with namespace aliases


Peeja:
## Expected behavior

Namespace form should be cleaned with no errors.
## Actual behavior

`clojure.lang.ExceptionInfo: Invalid token: ::foo/bar`

Sounds like `::foo/bar` is being evaluated without the `foo` alias in place.
## Steps to reproduce the problem

Run `cljr-clean-ns` in a namespace containing:

``` clj
(ns my-app.core
  (:require [foobar.core :as foo]))

::foo/bar
```
### clj-refactor.el version information

```
clj-refactor 2.3.0-SNAPSHOT (package: 20160524.146), refactor-nrepl 2.3.0-SNAPSHOT
```
### CIDER version information

```
;; CIDER 0.13.0snapshot (package: 20160524.714), nREPL 0.2.12
;; Clojure 1.8.0, Java 1.8.0_51
```
### Leiningen or Boot version

```
Leiningen 2.5.3 on Java 1.8.0_51 Java HotSpot(TM) 64-Bit Server VM
```
### Emacs version

24.4.1
### Operating system

OS X 10.9.5

expez:
You need to load the namespace into the repl or it won't be able to resolve the alias.  Once that is done it should work correctly.

Several people have been tripped up by this lately.  `clean-ns` is run so frequently I'm not sure prompting for a load would be a good either, though.  

Let me know if that does it for you.

Peeja:
Oh, I see. I can't do that: I'm in ClojureScript. Is there any way around that?

expez:
Should work in that way for cljs too.  

We might have a bug in cljs, then.

unkelly:
I ran into this problem trying to use `cljr-find-usages`. The offending instance is outside of the namespace I'm running in. I guess the project analysis isn't smart enough to load each namespace before analyzing?

unkelly:
Oh nice, running `cider-load-all-project-ns` prior to refactoring solves it 👍

Peeja:
@expez Huzzah! Thank you!

PR #396, opened 2017-11-07
Add an option to load all project namespaces at startup

joodie:
This ensures the AST cache is populated fully, speeding up
refactorings and prevents some issues when analizing qualified
keywords with aliased namespaces.

Enabling this option makes it unnecessary to manually call
`cider-load-all-project-ns` before running `cljr-find-usages`.

See also: #353

- [*] You've added tests (if possible) to cover your change(s)
   * I tried editing the find-usages.feature, but nothing I do there seems to do make any difference -  I cannot even make the existing tests fail...
  
- [X] You've updated the changelog (if adding/changing user-visible functionality)
- [*] You've updated the readme (if adding/changing user-visible functionality)
  * The related exiting features are all documented on the Wiki and not in the readme, so I've not changed the readme since this feature makes no sense to use on its own.

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## NEXT VERSION
+
+### Changes
+ - Add cljr-load-all-project-ns-on-startup option, to help prefill AST cache
+
 ## 2.3.1
 
 - [#363](https://github.com/clojure-emacs/clj-refactor.el/issues/363) cljr-favor-prefix-notation by default is set to false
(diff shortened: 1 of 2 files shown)


expez:
Hi, thanks for taking an interest!

The AST cache is its own thing an doesn't rely on REPL state.  

Unless `cljr-eagerly-build-asts-on-startup` is set to `nil` it will start building ASTs, in the background, the moment the REPL comes up.

PR #397, opened 2017-11-12
Let user specify default language context

### LINKED ISSUE #394: Request: variable to control cljr--clj-context-p popup
Hi, let me first say that I find clj-refactor incredibly helpful and make use of it every day. Thanks very much for such a great tool!

I have a fairly minor request: clj-refactor.el contains the following function:

```
(defun cljr--clj-context-p ()
  "Is point in a clj context?"
  (or (cljr--clj-file-p)
      (when (cljr--cljc-file-p)
        (if (cljr--point-in-reader-conditional-p)
            (cljr--point-in-reader-conditional-branch-p :clj)
          (string-equal (cljr--prompt-user-for "Language context at point? "
                                               (list "clj" "cljs"))
                        "clj")))))
```

Unlike pretty much everything else in clj-refactor, I find this more annoying than helpful; I nearly always escape out of the popup -- often after typing half the fn/var name I want, meaning I have to retype it once I realize I'm in the popup. I've taken a look to see if I saw a variable that controls this, but I haven't found one.

Can we have a user-settable variable that controls this, such that the popup is skipped if the variable is set to false? It could result in the same behavior that escaping out of the popup does (even if there's a bit of cost to clj-refactor's ability to help out in that case). Of course I'm imagining it set to true by default, so that there's no behavior change unless the user explicitly sets the variable in their user config.

I'm happy to try to put together a PR if that would be helpful.

Thanks!

eggsyntax:
When the user enters an apparent namespace like `somelib/`, and
the language context is ambiguous, clj-refactor creates a popup
asking the user to specify whether the context is "clj" or "cljs".
This change allows the user to specify a language context which
should be assumed, so that clj-refactor will choose that context
in ambiguous cases rather than creating a popup. Default behavior
is still to create the popup; this change only creates a variable
that lets the user override that behavior.
Tested fairly extensively at REPL; it didn't seem substantial
enough to warrant a new automated test (especially since I've
never worked with Cucumber).

Resolves #394 

** See above
** Note: it wasn't clear to me how to add it to the changelog, since the changelog is version-based, and this didn't seem to warrant a version bump. Happy to add it if this is clarified.
** README points to `https://github.com/clojure-emacs/clj-refactor.el/wiki#customization`, which just says how to pull up the set of available customizations (& this now comes up as part of that set).

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- [#394](https://github.com/clojure-emacs/clj-refactor.el/issues/394) New config option `cljr-assume-language-context`: by default, when clj-refactor encounters an ambiguous context (clj vs cljs) it creates a popup asking user which context is meant. If this option is changed to "clj" or "cljs", clj-refactor will use that as the assumed context in such ambigous cases.
+
 ## 2.3.1
 
 - [#363](https://github.com/clojure-emacs/clj-refactor.el/issues/363) cljr-favor-prefix-notation by default is set to false
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -221,6 +221,13 @@ won't run if there is a broken namespace in the project."
   :group 'cljr
   :type 'boolean)
 
+(defcustom cljr-assume-language-context nil
+  "If set to 'clj' or 'cljs', clj-refactor will use that value in situations
+  where the language context is ambiguous. If set to nil, a popup will be
+  created in each ambiguous case asking user to choose language context."
+  :group 'cljr
+  :type 'string)
+
 (defcustom cljr-libspec-whitelist
   '("^cljsns" "^slingshot.test" "^monger.joda-time" "^monger.json")
   "List of regexes to match against libspec names which shouldn't be pruned.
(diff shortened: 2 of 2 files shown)


expez:
Thanks for digging into this @eggsyntax! :+1: 

This looks mergeworthy to me but I'll give @benedekfazekas a chance to chime in.

Could you also create a changelog entry?  I'm sure other people might be interested in this too! :)

benedekfazekas:
no objections here!

eggsyntax:
@expez "Could you also create a changelog entry?"

Totally happy to! I was going to, but it wasn't clear to me how I should do so, since the changelog is version-based, and I assume this doesn't warrant a version bump. Should I start a new section above the 2.3.1 section?

expez:
> Should I start a new section above the 2.3.1 section?

Yes, something like [this](
https://github.com/clojure-emacs/refactor-nrepl/blob/master/CHANGELOG.md#unreleased)

eggsyntax:
Changelog updated. Thanks, y'all!

expez:
Fantastic work, @eggsyntax!  Thanks a lot! :+1:

eggsyntax:
My pleasure, thanks for accepting it :-)

PR #408, opened 2018-01-29
Introduce `cljr-building-ast-cache?`

vemv:
Allows one to build helpers to reload code at the right time, avoiding https://github.com/clojure-emacs/refactor-nrepl/issues/134

As this is a simple variable, not sure if should go into changelog/readme?

Cheers - Victor

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -276,6 +276,16 @@ if it appears to be unused."
 (defvar cljr--find-symbol-buffer "*cljr-find-usages*")
 (defvar cljr--post-command-messages nil "Message(s) to display after the current command is done.")
 
+(defcustom cljr-before-warming-ast-cache-hook nil
+  "Runs before each time the AST is loaded."
+  :group 'cljr
+  :type 'function)
+
+(defcustom cljr-after-warming-ast-cache-hook nil
+  "Runs after each time the AST is loaded."
+  :group 'cljr
+  :type 'function)
+
 ;;; Buffer Local Declarations
 
 ;; tracking state of find-symbol buffer
@@ -2670,9 +2680,13 @@ Also adds the alias prefix to all occurrences of public symbols in the namespace
                             asts-in-bad-state) "; "))))))
 
 (defun cljr--warm-ast-cache ()
+  (when cljr-before-warming-ast-cache-hook
+    (funcall cljr-before-warming-ast-cache-hook))
   (cljr--call-middleware-async
    (cljr--create-msg "warm-ast-cache")
    (lambda (res)
+     (when cljr-after-warming-ast-cache-hook
+       (funcall cljr-after-warming-ast-cache-hook res))
      (cljr--maybe-rethrow-error res)
      (cljr--maybe-nses-in-bad-state res)
      (when cljr--debug-mode
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - [#415](https://github.com/clojure-emacs/clj-refactor.el/issues/415) Support for deps.edn - based projects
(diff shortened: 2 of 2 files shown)


benedekfazekas:
perhaps a before and after hook would be much nicer and more useful too. would be easy to implement the above with them and more.

benedekfazekas:
and thanks for investing your time and energy :)

vemv:
✌️

Note that at least in the project I'm hacking on, analysis happens twice on startup:

image

So, for example, a callback that issued a `(init)` (or variations: `(go)`, etc. A dev facility to start Reloaded workflow) on AST-index-updated would be called twice, which would be undesirable.

Internally, the user-built helper would have to use internal variables. Which is kind of a roundabout way of doing what my proposed commit does.

I had planned to do this in my emacs.d:

```
(defun try-init ()
  (if cljr-building-ast-cache?
    (message "Try again in a bit")
    (clojure-eval "(repl/init)")))
```

i.e. the user may have to try the command a second time, which can be simpler than having callbacks, excess variables, etc

WDYT?

benedekfazekas:
I don't think it should happen twice. I can only try to reproduce in the evening tho

benedekfazekas:
hooks vs variable: I just think hooks are a more generic way for client code to interact with and also a superset of what you are proposing

clj-refactor.el
```diff
@@ -2614,11 +2616,13 @@ Also adds the alias prefix to all occurrences of public symbols in the namespace
                             asts-in-bad-state) "; "))))))
 
 (defun cljr--warm-ast-cache ()
+  (setq cljr-building-ast-cache? t)
   (cljr--call-middleware-async
    (cljr--create-msg "warm-ast-cache")
    (lambda (res)
      (cljr--maybe-rethrow-error res)
      (cljr--maybe-nses-in-bad-state res)
+     (setq cljr-building-ast-cache? nil)
```
expez: This should be the first thing, or it's going to report `t` because an error was rethrown, right?

vemv:
* Integrated PR feedback from both of you
* Added entry to the changelog
* Tested out with a sample functionality build on top of this. Works!
* The "analysis happens twice" thing was project-specific, due to nrepl connecting twice. Clearly unrelated; sorry for the noise

Finally, there's only one byte-compilation warning:

`Warning (bytecomp): Unused lexical argument ‘response’`

Dunno how to fix, also hesitant about improving it since you may dislike the empty lambdas (see diff) in the first place.

clj-refactor.el
```diff
@@ -269,6 +269,16 @@ if it appears to be unused."
 (defvar cljr--find-symbol-buffer "*cljr-find-usages*")
 (defvar cljr--post-command-messages nil "Message(s) to display after the current command is done.")
 
+(defcustom cljr-before-warming-ast-cache-hook (lambda ())
+  "Runs before each time the AST is loaded."
+  :group 'cljr
+  :type 'function)
+
+(defcustom cljr-after-warming-ast-cache-hook (lambda (response))
```
expez: Think it's customary to use `nil` as the default here, and then put the `funcall` below in a `when`.

I'm not complete sure, though, but I'd prefer something a little less elegant over a compile warning about unused locals.

benedekfazekas:
@vemv any progress on this? no rush, but as far as i can see there is a only a small step missing (see expez's comment)

vemv:
Hi, sorry I'm having a busy week! I'll do it asap

benedekfazekas:
no rush really. just did not want this to slip through the cracks

raxod502:
This is somewhat unrelated, but you can fix unused argument warnings by prepending the argument name with an underscore (to signify to the compiler that you intend the variable to be unused).

vemv:
Feedback addressed, branch rebased against master. Tested out again and the callbacks keep working.

expez:
Thanks for seeing this through, @vemv! :+1:

PR #424, opened 2018-07-02
Fix renamed cider function in resolve missing

tggreene:
This PR should update a function name that was updated in cider recently in this commit https://github.com/clojure-emacs/cider/commit/5cb98a74bad81a2cd15f8db86c469020ed030156

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -2847,7 +2847,7 @@ Date. -> Date
   ;; Just so this part can be mocked out in a step definition
   (when-let (candidates (thread-first (cljr--create-msg "resolve-missing"
                                                         "symbol" symbol
-                                                        "session" (cider-current-session))
+                                                        "session" (cider-nrepl-eval-session))
                           (cljr--call-middleware-sync
                            "candidates")))
     (edn-read candidates)))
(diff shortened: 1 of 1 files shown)


expez:
Nice, thanks! :+1: 

Want to grab the other one, that's causing the build to fail, while you're in there?

bbatsov:
Keep in mind this will break clj-refactor for the current stable cider
release.

tggreene:
Have just tried to determine what's happening with the failing build and appears to affect master HEAD:

Head of clj-refactor.el [632be9e](https://github.com/tggreene/clj-refactor.el/commit/632be9e9bab1045fd5c1dcacd4f5781c2a49db81) [my branch to test build](https://travis-ci.org/tggreene/clj-refactor.el/builds/399592656)

I'm a bit out of my depth with emacs lisp but will have another look this evening.

bbatsov:
Btw, I just noticed that both this functions have obsoletions aliases in CIDER's `master`, which is another point in favor of not renaming them here yet. It seems nothing should be broken because of the names, likely the breakage is coming from the changes in the file structure of CIDER (e.g. those functions are now in `cider-connection.el`.

tggreene:
Indeed, seems sensible. I was working on a slightly older cut! I'll close this PR

PR #427, opened 2018-07-08
[Fix #426] Expose variable to ignore paths in the middleware

### LINKED ISSUE #426: Building the AST is not practical because it runs tests - how to prevent that?
`cljr-find-usages` and friends have to create the AST for the whole project.

This is not practical for my use case, as it would need to run all the tests, which might take a long time. Also, some tests actually break the AST creation with `Method code too large!`.

Is there are way to configure `clj-refactor` so that it ignore certain namespaces when building the AST?

expez:
This makes it possible to easily exclude certain files and directories
for analysis.

The code for this already exists in the middleware.  Not sure why a setting was never created.  Perhaps we just forgot?

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -286,6 +286,12 @@ if it appears to be unused."
   :group 'cljr
   :type 'hook)
 
+(defcustom cljr-middleware-ignored-paths nil
+  "List of (Java style) regexes to paths that should be ignored
+  by the middleware."
+  :group 'cljr
+  :type '(repeat string))
+
 ;;; Buffer Local Declarations
 
 ;; tracking state of find-symbol buffer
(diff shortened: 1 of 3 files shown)


expez:
@bbatsov While I'm in here, any idea about how to fix the build?

bbatsov:
@expez I see that the build is failing because CIDER's master now depends on Emacs 25.1 (as `sesman`, the new connection management library depends on it). I guess you'll have to update the build matrix.

bbatsov:
> clj-refactor.el:289:1:Error: defcustom for ‘cljr-middleware-ignore-paths’ fails to specify containing group

That fails because the group of the `defcustom` is not specified. I guess you didn't configure the top-level group to just infer that everything starting with `cljr` should go to the same group.

clj-refactor.el
```diff
@@ -2484,6 +2488,7 @@ See: https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-promote-function
                             "line" line
                             "column" column
                             "name" symbol
+                            "ignore-paths" cljr-middleware-ignore-paths
```
bbatsov: I wonder if `ignored-paths` won't be a better name for this, as normally I'd assume something named `ignore-paths` means something else.

expez:
@bbatsov I renamed the variable in clj-refactor, but the variable name in the backend was already commited to when we released 2.3.1.

expez:
Merged manually

PR #431, opened 2018-12-23
Add 'back' commands to the Hydra children menus

vemv:
This seems handy for exploration, e.g. I don't know what command I'm looking for, so I tentatively choose sub-hydra _x_, then I need the ability to go back.

Tried out, works as intended!


expez:
Great, thanks! 👍

PR #439, opened 2019-05-16
[Fix #402] Do not alter strings when using `cljr-stop-referring`.

### LINKED ISSUE #402: cljr-stop-referring alters string literals
Hey there,

If `foo` was a referred symbol from `some-ns`, and the current buffer contains a literal string such as `"Hello foo!"`, s-r will turn the string into `"Hello some-ns/foo!"`.

Cheers - Victor

anthonygalea:


magnars:
Excellent, thank you!

PR #440, opened 2019-05-17
[Resolve #366] Add `cljr-rename-ns-alias`.

### LINKED ISSUE #366: Support for renaming ns aliases
Especially with new syntax in Clojure 1.9 for making namespaced maps easier to use, including e.g. `#::x{:a 1}` and `{::x/keys [a b c]}`, this is becoming a more and more useful feature that's tricky to write the correct regex alternative for.

anthonygalea:
Refactoring works but I have an issue with the `Without :as` scenario. I think it's related to the action chain but not sure how to fix it. My goal there is to ensure that if a rename is made for an alias which is not in the `require` the file is untouched.


anthonygalea:
Ok ignore my last comment. I see I was wrongly using `user-error` instead of `message`.

expez:
Hi, thanks for taking an interest @anthonygalea!  This is great stuff! 👍 

Since this is elisp only, do you think we should put this in clojure-mode @bbatsov?

anthonygalea:
Thanks @expez, @bbatsov if this should go into clojure-mode just let me know, happy to look into it.

bbatsov:
@anthonygalea Sorry about the slow response! Let's add this directly to `clojure-mode`!

anthonygalea:
No problem @bbatsov. Opened another [pull request](https://github.com/clojure-emacs/clojure-mode/pull/529) in `clojure-mode` so I'm closing this one.

PR #450, opened 2019-10-31
Implement `cljr-magic-use-namespaces`

jgertm:
This feature is similar to `cljr-magic-require-namespaces` but instead
of creating aliases, it refers all symbols from the namespace.

I have a hunch that this would be better if it just exposed the list option and triggered the logic when the list is non-empty. [Any thoughts?](https://github.com/clojure-emacs/clj-refactor.el/pull/450/commits/6e8f829f2a420f0a5586fb4f59026f1cf7562d42)

---

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- [#450](https://github.com/clojure-emacs/clj-refactor.el/pull/450) New config options: `cljr-magic-uses` (boolean) and `cljr-magic-uses-namespaces` (list): if boolean is `true`, upon creation of new namespace, add `:refer :all` entries for all namespaces in the list to the new namespace
+- [#450](https://github.com/clojure-emacs/clj-refactor.el/pull/450) New config option: `cljr-magic-uses-namespaces` (list): if list is non-empty, upon creation of new namespace, add `:refer :all` entries for all namespaces in the list to the new namespace
 - [#402](https://github.com/clojure-emacs/clj-refactor.el/issues/402) cljr-stop-referring: do not alter strings.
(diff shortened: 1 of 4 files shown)


expez:
Sorry for letting this languish for so long.  I had some thoughts that I wanted to let bake a little before I replied, and then forgot all about it 😞 

First of all, I really dislike `:refer :all` and feel very strongly that we (as builders of tooling) should limit ourselves to making the *right* things easy.  This gently steers people into adopting best-practices and makes life better for everyone in the long run.

My second thought builds on 1) but addresses the particular implementation chosen for this feature.  By adding a list of namespaces to refer wholesale I fear that the path of least resistance is to add tons of stuff to this list.

We already have a feature along this path, `cljr-add-use-to-ns`.  Perhaps we could extend this to, when given a prefix argument, show you a completing read of the list in a variable like the one you proposed?

jgertm:
Sorry for letting your reply languish even longer.

I agree that, in general, `:refer :all` is to be avoided. 
However, I believe that every project gets to have one exception: a `$PROJECT.prelude` ns that contains non-domain specific amendments to Clojure's standard library. I usually put stuff like `clojure.core.match/match` and `when-let+` (like `when-let`, but with multiple bindings) in there. This namespace should be available anywhere in the project, so that's why I implemented this feature. If you feel that is sufficient justification to add it to clj-refactor, I'm happy to clean this up. If not, feel free to close the PR.

vemv:
This doesn't sound like a feature that we'd like to encourage as it's a practice that has fallen out of favor over the last years. You can find a trace of that thinking in the clj-kondo and Eastwood linters, in https://github.com/bbatsov/clojure-style-guide, and https://stuartsierra.com/2016/08/27/how-to-ns which was a very influential article.

Emacs being Emacs, probably you'd have it reasonably easy to cleanly implement this feature without altering clj-refactor.el source code? e.g. by using `defadvice`

vemv:
> This doesn't sound like a feature that we'd like to encourage as it's a practice that has fallen out of favor over the last years.

PR #457, opened 2019-12-15
[Fix #456] Add ignore-paths to warm-ast-cache message

### LINKED ISSUE #456: AST cache warm-up process ignores middleware ignored-paths
## Expected behavior
AST cache warm-up doesn't analyze (nor evaluate) paths added to `cljr-middleware-ignored-paths`.

## Actual behavior
Warm-up goes through all directories in classpaths.

A bit of technicals:

Middleware functions base on `dirs-on-classpath` function to calculate list of directories (and subsequently files) to analyze. This functions also filters out all the directories matching patterns listed in `:ignore-paths` configuration.

https://github.com/clojure-emacs/refactor-nrepl/blob/c23c00b326816b593df8279e500d2ecb23d433ba/src/refactor_nrepl/core.clj#L54

By default `:ignore-paths` list is empty, but can be easily overriden with each message that clj-refactor sends with "ignore-paths" option, just like here:

https://github.com/clojure-emacs/clj-refactor.el/blob/50d2d8aad5e0bd8002173b300f8419d72ceab7af/clj-refactor.el#L2515

The thing is that `cljr--warm-ast-cache` doesn't send this option which forces middleware to calculate AST cache for all classpath entries, even though some of them won't be ever used:

https://github.com/clojure-emacs/clj-refactor.el/blob/50d2d8aad5e0bd8002173b300f8419d72ceab7af/clj-refactor.el#L2714

In practice, when I set the `cljr-middleware-ignored-paths` to ignore tests, the AST cache will be built for all tests anyway, although for example `find-symbol` will never reach for these cached AST nodes (as it honors ignored-paths).

That also might be the reason why this problem wasn't probably entirely solved:
https://github.com/clojure-emacs/clj-refactor.el/issues/426

mbuczko:
This change adds `ignore-paths` to `warm-ast-cache` to be sure cache is not created (and corresponding files are not analyzed nor evaluated) for ignored directories.

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -2711,7 +2711,8 @@ Also adds the alias prefix to all occurrences of public symbols in the namespace
 (defun cljr--warm-ast-cache ()
   (run-hooks 'cljr-before-warming-ast-cache-hook)
   (cljr--call-middleware-async
-   (cljr--create-msg "warm-ast-cache")
+   (cljr--create-msg "warm-ast-cache"
+                     "ignore-paths" cljr-middleware-ignored-paths)
    (lambda (res)
      (run-hook-with-args 'cljr-after-warming-ast-cache-hook res)
      (cljr--maybe-rethrow-error res)


expez:
Thanks @mbuczko!  

Any interest in digging into the build failure?  We're building against a very old CIDER version, so perhaps just bumping that'll solve it...

mbuczko:
sure, I will take a look at the building problem. thanks for merging this change.

PR #458, opened 2019-12-17
Update cider dependency and GNU ELPA gpg keys

mbuczko:
The problem of failing builds is mainly caused by expired elpa gpg keys. I also updated cider dependency, although I'm neither sure we need to do that, nor I don't really know what version should we depend on? snapshot? :)

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -8,7 +8,7 @@
 ;;         Benedek Fazekas <benedek.fazekas@gmail.com>
 ;; Version: 2.5.0-snapshot
 ;; Keywords: convenience, clojure, cider
-;; Package-Requires: ((emacs "25.1") (seq "2.19") (yasnippet "0.6.1") (paredit "24") (multiple-cursors "1.2.2") (clojure-mode "5.6.1") (cider "0.17.0") (edn "1.1.2") (inflections "2.3") (hydra "0.13.2"))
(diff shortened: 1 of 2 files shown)


clj-refactor.el
```diff
@@ -8,7 +8,7 @@
 ;;         Benedek Fazekas <benedek.fazekas@gmail.com>
 ;; Version: 2.5.0-snapshot
 ;; Keywords: convenience, clojure, cider
-;; Package-Requires: ((emacs "25.1") (seq "2.19") (yasnippet "0.6.1") (paredit "24") (multiple-cursors "1.2.2") (clojure-mode "5.6.1") (cider "0.17.0") (edn "1.1.2") (inflections "2.3") (hydra "0.13.2"))
+;; Package-Requires: ((emacs "25.1") (seq "2.19") (yasnippet "0.6.1") (paredit "24") (multiple-cursors "1.2.2") (clojure-mode "5.6.1") (cider "0.24.0-snapshot") (edn "1.1.2") (inflections "2.3") (hydra "0.13.2"))
```
mbuczko: I'm not sure what cider version should we depend on here?

bbatsov: Generally it should the latest stable CIDER release.

expez: I agree with @bbatsov that we should change this to be the latest stable release, and aim to update it more frequently than we've done in the past.

If we depend on the snapshot we're giving ourselves very little lead-time to react to the occasional breaking change in CIDER.

mbuczko:
ok, changed cider to 0.23.0. this is the latest stable, isn't it?

bbatsov:
Thanks!

PR #463, opened 2020-02-20
Fix require to use parseedn feature instead of edn feature

### LINKED ISSUE #462: Incorrect Require for parseedn Package in clj-refactor.el
Current version of clj-refactor makes a `require` statement which depends on the `edn` elisp feature, which is no longer provided by any of the dependencies of clj-refactor. The new feature provided is `parseedn`, and the functions from that version have been used, however the feature require has not been upgraded. This causes many problems in clojure files since it stops processing of many additional hooks on clojure files, waiting for the require to be corrected.

## Expected behavior
clj-refactor requires the correct feature name.

## Actual behavior
clj-refactor requires `edn` when the provided feature is `parseedn`

Run the following command with either `$HOME` set to a temporary value for the command to allow a clean location for .emacs.d, or have a clean .emacs.d.
```
emacs -Q --eval "(progn (require 'package) (add-to-list 'package-archives '(\"melpa\" . \"http://melpa.org/packages/\") t) (package-initialize) (package-refresh-contents) (package-install 'clj-refactor))"
```
Watch the window which is created, and when clj-refactor begins to compile, an error like the following will appear:
```

Compiling no file at Thu Feb 20 16:26:29 2020
Leaving directory ‘/home/suskeyhose/tmp/.emacs.d/elpa/clj-refactor-20200219.1038’

Compiling file /home/suskeyhose/tmp/.emacs.d/elpa/clj-refactor-20200219.1038/clj-refactor-compat.el at Thu Feb 20 16:26:29 2020
Entering directory ‘/home/suskeyhose/tmp/.emacs.d/elpa/clj-refactor-20200219.1038/’

Compiling file /home/suskeyhose/tmp/.emacs.d/elpa/clj-refactor-20200219.1038/clj-refactor.el at Thu Feb 20 16:26:29 2020
clj-refactor.el:46:1:Error: Cannot open load file: No such file or directory, edn
```

According to ELPA, the latest version they have is 20200219.1038. cljr-version reports version `2.5.0-SNAPSHOT`.
This problem does not occur on stable.

### Emacs version
26.3

### Operating system
Arch Linux

## Fix
This fix for this should be incredibly simple based on my testing, the only change needed is to change line 47 of clj-refactor.el to the following:
```lisp
(require 'parseedn)
```

IGJoshua:

- [N/A] The commits are consistent with our [contribution guidelines](CONTRIBUTING.md)
- [N/A] You've added tests (if possible) to cover your change(s)
- [N/A] You've updated the changelog (if adding/changing user-visible functionality)
- [N/A] You've updated the readme (if adding/changing user-visible functionality)

This fixes an error with loading clj-refactor.el on installs that don't already have the `edn` package available. The `edn` package is no longer required by clj-refactor.el (being replaced by `parseedn`), but still makes the incorrect require call.

The only test which isn't passing, also wasn't passing before the change, it being the circular-dependency check.

Fixes #462

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -44,7 +44,7 @@
 (require 'multiple-cursors-core)
 (require 'clojure-mode)
 (require 'cider)
-(require 'edn)
+(require 'parseedn)
 (require 'sgml-mode)
 (require 'inflections)
 (require 'hydra)


bbatsov:
Thanks!

Issue #467, opened 2020-04-04
cljr-move-file freezes without a clue of what it's doing


jimka2001:
#465  Expected behavior
When I use M-x cljr-move-form, i expect to see something like what happens on 
[https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-move-form](the move form video).
It should ask me which file I'd like to move the top level from into.

## Actual behavior

When I put the cursor inside a form and use M-x cljr-move-form, emacs seems to freeze, nothing happens.  If I press C-g, the cursor comes back but the form containing the cursor has vanished.  i can get it back with C-y, so no harm done.

The same thing happens if I select several forms, and then M-x cljr-move-form.

## Steps to reproduce the problem

Put the cursor inside a form and press M-x cljr-move-form.

I'm providing the project I'm using.  The form I tried it in was
in the file `clojure-rte/src/clojure_rte`.  My cursor was sitting on line 560.

[clojure-rte.zip](https://github.com/clojure-emacs/clj-refactor.el/files/4431176/clojure-rte.zip)

### clj-refactor.el version information
"clj-refactor 2.5.0 (package: 20200229.1105), refactor-nrepl is unreachable"

### CIDER version information

;; Connected to nREPL server - nrepl://localhost:63956
;; CIDER 0.24.0 (India), nREPL 0.6.0
;; Clojure 1.10.0, Java 14

### Leiningen or Boot version
Leiningen 2.9.3 on Java 14 Java HotSpot(TM) 64-Bit Server VM

### Emacs version
GNU Emacs 26.3 (build 1, x86_64-apple-darwin18.2.0, NS appkit-1671.20 Version 10.14.3 (Build 18D109)) of 2019-09-02

### Operating system
MacOS Catalina
Version 10.15.3
MacBook Pro (13-inch, 2018, Four Thunderbolt 3 Ports)

expez:
Was this solved by #468 @jimka2001?

PR #468, opened 2020-04-04
Balance parentheses

prertik:

- [ ] The new code is not generating byte compile warnings (run `cask exec emacs -batch -Q -L . -eval "(progn (setq byte-compile-error-on-warn t) (batch-byte-compile))" clj-refactor.el`)

Thanks!


prertik:
Before, applying my commit, I noticed the tests weren't running because of unbalanced parentheses.
Now, the tests fail at: `Scenario: Move forms does not create circular dependencies, #341`. I think this is related to issue #467. @bbatsov

bbatsov:
Ops, my bad! Thanks for fixing my mistake! That's what happens when you just commit away without testing at all. :D My guess is that #467 was actually caused by the broken code, we'll see about that.

PR #473, opened 2020-07-18
Mention the name of the middleware namespace in README

blak3mill3r:
For users who are not using cider-jack-in or leiningen, I think it's nice to have the middleware namespace mentioned here.


bbatsov:
Agreed. Thanks!

Issue #442, opened 2019-06-02
cljr-clean-ns easy fix


boymaas:
Excuses no time to fill in the complete form. Just a quick pointer, got the following error `string-prefix-p: Wrong type argument: stringp, nil` when trying to use `cljr-clean-ns`. A quick debug shows a very quick fix.

https://github.com/clojure-emacs/clj-refactor.el/blob/9f508bbf7bf8dd439a78066a9d6fa47f4bc8f64f/clj-refactor.el#L2738-L2740

It uses a `let` which I believe should be a `let*`. Wondering why this was not noticed before.

Thanks for a great library!!

expez:
Thanks for the report!

> Wondering why this was not noticed before.

This code path was added about a month ago and is used only in some situations.

expez:
[solved](https://github.com/clojure-emacs/clj-refactor.el/pull/474)

PR #474, opened 2020-08-27
Send nREPL-based paths instead of CIDER-based paths

iarenaza:
When `cider-path-translations` is in use (e.g., when CIDER is running
on the host, but the nREPL is running in a VM or a Docker container)
we need to translate the local CIDER-based paths to the remote
nREPL-based paths for those operations that send paths to nREPL.

We use the new `cider-to-nrepl-filename-function` that is introduced
in CIDER pull request #2897, that should be applied before this one.

Also includes a fix for #442.

Thanks!


expez:
Excellent!  Thank you, @iarenaza!

PR #484, opened 2021-03-19
Fix clojure-cli warning about unqualified dependencies

smile13241324:
Without this running `cider-jack-in` in a clojure project
with clojure-cli will produce warnings about unqualified
dependencies which may cause errors in future releases.

This PR makes sure that all injected dependencies are
qualified.

- [-] You've added tests (if possible) to cover your change(s)
- [-] All tests are passing (run `./run-tests.sh`) -> one test is failing but this is also happening on master and is most likely be caused by my system.
- [-] You've updated the changelog (if adding/changing user-visible functionality) -> Functionality is not user-visible
- [-] You've updated the readme (if adding/changing user-visible functionality) -> Functionality is not user-visible


lvh:
No merge privileges, but this looks great to me :-) (And I'm affected by the bug in the first place :D)

expez:
Thanks for the PR @smile13241324.  Much appreciated! 👍

PR #486, opened 2021-04-11
Add a step to upgrade cask

vedang:
This is a (potential) fix for the tests failing against Emacs 25. I
used this config file as a template to create the config file for
vedang/pdf-tools@6ddfda8b and faced the same problem. The failure was
in Cask not being able to install `cl-lib`, which seemed strange to
me. I guessed that the emacs-25 image might not have been baked
properly.

Adding this step fixed the problem for my repository.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,6 +5,7 @@ default: &default-steps
   steps:
     - checkout
     - run: apt-get update && apt-get install make
+    - run: cask upgrade-cask || true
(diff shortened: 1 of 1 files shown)


bbatsov:
Thanks for looking into this! //cc @Silex

Silex:
Ah, if the fix is that simple I'll try to incorporate it, but I suspect it'll go away anyway once I switch to purcell's emacs.

PR #489, opened 2021-05-02
Make cljr-slash a no-op on fractions

### LINKED ISSUE #481: cljr-slash documentation seems out of date
## Expected behavior

Docstring for `cljr-slash` correctly describes behavior of the function

## Actual behavior

Docstring says 
```
If ‘cljr-magic-require-namespaces’ is non-nil, typing one of the
short aliases listed in ‘cljr-magic-requires’ followed by this
command will add the corresponding require statement to the ns
form.
```

However, having done no customization:
- cljr-magic-requires is `t`
- cljr-magic-require-namespaces is 
```
(("io" . "clojure.java.io")
 ("set" . "clojure.set")
 ("str" . "clojure.string")
 ("walk" . "clojure.walk")
 ("zip" . "clojure.zip"))
```

So it seems the docstring has confused which var is the list and which is the toggle. 

Also, cljr-magic-require-namespaces is a very short list, but cljr-slash works for many other aliases that I use in my project, so the claim that cljr-slash just pulls from a list does not seem true either.

clj-refactor 2.5.1, refactor-nrepl 2.5.0

### CIDER version information

CIDER 1.0.0 (package: 20210125.1630) (Sofia), nREPL 0.8.0
Clojure 1.10.1, Java 11.0.8

### Emacs version

GNU Emacs 27.1 (build 1, x86_64-redhat-linux-gnu, GTK+ Version 3.24.21, cairo version 1.16.0) of 2020-08-20

### Operating system

Fedora release 32 (Thirty Two)

yuhan0:
This prevents the expensive alias lookup when typing the slash in fraction literals eg. `1/2`.
Also updates the docstring, fixes #481.

There is one unrelated failing test involving to cljr-move-form, which also fails on the master branch.

----

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- [#483](https://github.com/clojure-emacs/clj-refactor.el/issues/483): Improve performance of cljr-slash when typing fraction literals.
 - [#482](https://github.com/clojure-emacs/clj-refactor.el/issues/482): Add missing defgroup form.
 - [#470](https://github.com/clojure-emacs/clj-refactor.el/issues/470): Choose `deps.edn` over `pom.xml` as project file.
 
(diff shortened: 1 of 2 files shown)


bbatsov:
Thanks!

Issue #492, opened 2021-07-08
cljr-slash: rule out some edge cases


vemv:
For cljr-slash functionality, some edge cases will predictably never succeed, yet will spend computational time.

e.g. `a/a/` has two slashes, there will be nothing matching that.

Similarly sometimes as I use emacs I make other typos that also can never succeed, yet block the UI for a few secs.

So one could have a simple regex for validating the expected admissible inputs.

Issue #491, opened 2021-06-28
Fix the build


expez:
Looks like two things are going on:

1. One actual test failure for emacs versions 26 and 27.
2. Test check validating grammar etc

Can we turn off test check?  I'd rather have a green build badge giving an indication that this lib is alive and kicking.
Furthermore, with limited time available I'd much rather fix actual bugs than go over 3k lines of code to make sure e.g. error messages don't end in a period.

bbatsov:
I'd disable the offending test as the quickest solution. I'm also fine with disabling `checkdoc`, although it's not grammar check. Still, it's not something particularly important.

bbatsov:
Btw, when I checked the CI I saw that only the one test if failing.

vemv:
> I'd disable the offending test as the quickest solution. 

(We did so)

PR #493, opened 2021-07-08
Introduce `completable-for-cljr-slash?`

vemv:
Fixes https://github.com/clojure-emacs/clj-refactor.el/issues/492


clj-refactor.el
```diff
@@ -1983,37 +1983,68 @@ the alias in the project."
     (backward-sexp 1)
     (looking-at-p "[-+0-9]")))
 
+(defun completable-for-cljr-slash? (sym)
+  (when sym
+    (not (null (string-match-p "^\\([a-zA-Z]+[a-zA-Z0-9\\-]*\\)+\\(\\.?[a-zA-Z]+[a-zA-Z0-9\\-]*\\)*$" sym)))))
+
+(progn
```
vemv, line 1990: Quite obviously I don't intend to ship `assert`s, but seemingly there weren't unit tests in the test suite.

I do appreciate the `features` but I'm unfamiliar with their Emacs variant. More importantly this seems the perfect thing that can be tested directly in terms of inputs->outputs for a specific, pure `defun`.

If unit tests SGTY, I'd appreciate some guidance on how to actually ship them into CI.

expez: I've used ERT in the past, but CIDER uses [buttercup](https://github.com/jorgenschaefer/emacs-buttercup) which addresses some of the shortcomings of ERT.

If you're up for it, it would be great to use the same for clj-refactor. If so, we could probably just copy the setup that's in use in CIDER. 

Does that sound like a good idea @bbatsov or do you have some reservations about `buttercup` after using it in CIDER?

@vemv the current test suit in clj-refactor is written in [ecukes](https://github.com/ecukes/ecukes) and we've been wanting to get away from that for some time, since it stopped being maintained.  As I look up the link now I see that it's been moved to its own org and seems to have acquired a new maintainer.  Perhaps there's some hope in getting our test suite back into green...

I have mixed feelings about ecukes.  In general it's a really good fit, but in practice it has been quite finicky to work with.  It would be a lot of work to rewrite the features, though, so if we can update our ecukes version and get them running again that would probably be the best solution.

expez (changes requested):
Great idea for an improvement 👍

clj-refactor.el
bbatsov: Buttercup is awesome. No reservations about it at all.

clj-refactor.el
```diff
@@ -1983,37 +1983,68 @@ the alias in the project."
...
+  (assert (not (completable-for-cljr-slash? nil)))
+  (assert (completable-for-cljr-slash? "a"))
+  (assert (completable-for-cljr-slash? "a-"))
+  (assert (completable-for-cljr-slash? "a2"))
+  (assert (completable-for-cljr-slash? "a-2"))
+  (assert (completable-for-cljr-slash? "a2-"))
+  (assert (completable-for-cljr-slash? "a2.a"))
+  (assert (completable-for-cljr-slash? "a2.a-"))
+  (assert (completable-for-cljr-slash? "a2.a2"))
+  (assert (completable-for-cljr-slash? "a2.a-2"))
```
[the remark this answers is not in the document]
vemv: Note to self: also cover metadata

clj-refactor.el
```diff
@@ -1983,37 +1983,68 @@ the alias in the project."
...
-                       (yes-or-no-p (format "Add %s :as %s to requires?" long short))))
-          (save-excursion
-            (cljr--insert-in-ns ":require")
-            (let ((libspec (format "[%s :as %s]" long short)))
-              (insert libspec)
-              (ignore-errors (cljr--maybe-eval-ns-form))
-              (cljr--indent-defun)
-              (cljr--post-command-message "Required %s" libspec))))))))
+  (let ((original-input (cider-symbol-at-point)))
+    (insert "/")
```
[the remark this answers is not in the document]
vemv: if we insert the `/` before anything else, the `cider-symbol-at-point` (`thing-at-point` now) query will be affected by that mutation. So it will be harder to get an accurate answer

clj-refactor.el
```diff
@@ -1983,37 +1983,68 @@ the alias in the project."
...
+                            (completable-for-cljr-slash? original-input)
+                            (not (cljr--in-map-destructuring?))
+                            (not (cljr--in-ns-above-point-p))
+                            (not (cljr--in-reader-literal-p))
+                            (not (cider-in-comment-p))
+                            (not (cider-in-string-p))
+                            (not (cljr--in-keyword-sans-alias-p))
+                            (not (cljr--in-number-p))
+                            (clojure-find-ns)
+                            (cljr--magic-requires-lookup-alias)))
```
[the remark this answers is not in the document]
vemv: > Btw, I always program on a desktop computer so this op is never slow for me, 

The story is probably more complex b/c I have a 16 core machine and cljr-slash takes 7s if not parallelized. So it can depend on "mystery factors" :)

Most of the slowness comes from refactor-nrepl. This PR is concerned with preventing refactor-nrepl from being queried at all - it's always  a nice improvement even when the backend is cached and parallelized.

> but one more thing we could do is to move the op from sync to async. 

This could plausibly create a race condition: one could press `/`, keep typing, get one's input inserted _and then_ be offered completions.

> One thing I often do that will get affected by this is type obscure-ns/a^H which will then pop up the completion candidates for that ns and I can then search for the thing I barely remember.

Mind to expand? Not sure I follow

vemv (commented):
Thanks guys for the guidance!

CI is looking decent. However it might be a good idea to prioritize https://github.com/clojure-emacs/clj-refactor.el/issues/491 before proceeding with this PR.

clj-refactor.el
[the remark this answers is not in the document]
expez: Yeah, but we've had something that's worked pretty well for years, that we can use instead of `thing-at-point`:

```el
(thread-last (buffer-substring-no-properties
                             (cljr--point-after 'paredit-backward)
                             (1- (point)))
                 (string-remove-prefix "::")
                 (string-remove-prefix "@"))
```

Though looking at this, after you mentioned metadata, I see that it might not handle that properly.

clj-refactor.el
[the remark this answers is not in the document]
expez: > Most of the slowness comes from refactor-nrepl. This PR is concerned with preventing refactor-nrepl from being queried at all - it's always a nice improvement even when the backend is cached and parallelized.

Definitely agree with that!  I was just thinking out loud about other ways we might bring the 7s down or avoid just sitting there waiting.

> This could plausibly create a race condition: one could press /, keep typing, get one's input inserted and then be offered completions.

Yeah that's a fair point, and I'm not sure what's worse: sitting around waiting for 7 seconds, or getting a prompt to narrow down a selection of candidates from a `/` you typed 7 seconds ago.  In my own projects I very rarely see the prompt because I'm really strict about consistent aliases.

> Mind to expand? Not sure I follow

Let's say I'm looking to make a `queue` and I remember that `medley` has a constructor, but I can't remember if it's `->queue` or simply `queue`.  So I type `medley/` and the lib is required and the ns form gets re-evaluated so all the symbols are available to the completion engine.  Now I just need to trigger the list of every symbol in that ns, so I type say `q` but I'm not sure the thing I'm looking for starts with `q` so I erase that letter and now I can hit `C-s` to search among the completion candidates.  If I search for `queue` I'd see both `queue` and `->queue`.
The problem is that none of this will work if the namespace hasn't yet been added to the ns form and re-evaluated.

Typing this all out made me realize that I've also been working around a weird quirk that `/` doesn't trigger completions on it's own, yet it's perfectly happy with displaying that same if I type one char and then remove it...
Does it work that way for you?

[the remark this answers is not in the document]
expez: > Most of the slowness comes from refactor-nrepl. This PR is concerned with preventing refactor-nrepl from being queried at all - it's always a nice improvement even when the backend is cached and parallelized.

Definitely agree with that!  I was just thinking out loud about other ways we might bring the 7s down or avoid just sitting there waiting.

> This could plausibly create a race condition: one could press /, keep typing, get one's input inserted and then be offered completions.

Yeah that's a fair point, and I'm not sure what's worse: sitting around waiting for 7 seconds, or getting a prompt to narrow down a selection of candidates from a `/` you typed 7 seconds ago.  Although, in my own projects I very rarely see the prompt because I'm really strict about consistent aliases.

> Mind to expand? Not sure I follow

Let's say I'm looking to make a `queue` and I remember that `medley` has a constructor, but I can't remember if it's `->queue` or simply `queue`.  So I type `medley/` and the lib is required and the ns form gets re-evaluated so all the symbols are available to the completion engine.  Now I just need to trigger the list of every symbol in that ns, so I type say `q` but I'm not sure the thing I'm looking for starts with `q` so I erase that letter and now I can hit `C-s` to search among the completion candidates.  If I search for `queue` I'd see both `queue` and `->queue`.
The problem is that none of this will work if the namespace hasn't yet been added to the ns form and re-evaluated.

Typing this all out made me realize that I've also been working around a weird quirk that `/` doesn't trigger completions on it's own, yet it's perfectly happy with displaying that same if I type one char and then remove it...
Does it work that way for you?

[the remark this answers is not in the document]
vemv: > Yeah that's a fair point, and I'm not sure what's worse: sitting around waiting for 7 seconds, or getting a prompt to narrow down a selection of candidates from a / you typed 7 seconds ago.

The 7 seconds won't be a thing anymore following https://github.com/clojure-emacs/refactor-nrepl/pull/320 (and https://github.com/clojure-emacs/refactor-nrepl/issues/230#issuecomment-876645429 for cljs)

So both experiences (async vs. sync) would be as quick. But sync would avoid race conditions

>  Now I just need to trigger the list of every symbol in that ns, so I type say q but I'm not sure the thing I'm looking for starts with q so I erase that letter and now I can hit C-s to search among the completion candidates.

Personally I use the company-mode integration, so that I can see all names for a given prefix without having to type  a few chars tentatively:

image

I have to type `/ TAB`, so that's two different commands being composed. (I think that answer one of the questions)

Overall the cljr-slash integration works well for me - it successfully inserts a prefix and reloads the ns form

clj-refactor.el
```diff
@@ -1983,37 +1983,68 @@ the alias in the project."
...
-                       (not (> (length (cl-second aliases)) 1)) ; already prompted
-                       (yes-or-no-p (format "Add %s :as %s to requires?" long short))))
-          (save-excursion
-            (cljr--insert-in-ns ":require")
-            (let ((libspec (format "[%s :as %s]" long short)))
-              (insert libspec)
-              (ignore-errors (cljr--maybe-eval-ns-form))
-              (cljr--indent-defun)
-              (cljr--post-command-message "Required %s" libspec))))))))
+  (let ((original-input (cider-symbol-at-point)))
```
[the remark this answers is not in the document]
vemv: Great call, changed to `(thing-at-point 'symbol)`

clj-refactor.el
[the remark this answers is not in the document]
vemv: `thing-at-point` is working nicely, thanks!

The code you quote is also useful for the next step, I refined it in one of the commits (`Extract and complete cljr--namespace-of-thing-at-point`)

vemv:
Ready for re-review

clj-refactor.el
```diff
@@ -1983,37 +1988,47 @@ the alias in the project."
     (backward-sexp 1)
     (looking-at-p "[-+0-9]")))
 
+(defun completable-for-cljr-slash? (sym)
```
bbatsov, line 1991: All definitions should have a docstring, especially the "public" ones.

clj-refactor.el
```diff
@@ -1983,37 +1988,47 @@ the alias in the project."
...
+  (when sym
+    (or
+     ;; vanilla symbols, maybe prefixed by ^, :: or ^::
+     (not (null (string-match-p "^\\^?\\(::\\)?\\([a-zA-Z]+[a-zA-Z0-9\\-]*\\)+\\(\\.?[a-zA-Z]+[a-zA-Z0-9\\-]*\\)*$" sym)))
```
bbatsov, line 1995: Seems to me you can rewrite this code with the not null check and it'd read better.

vemv: I cargo-culted the pattern from https://github.com/magnars/s.el/blob/08661efb075d1c6b4fa812184c1e5e90c08795a9/s.el#L281-L283 (`s--truthy?` is used a few times there, so transitively the not-null check is used a bunch of times there)

No strong opinion from my side, but maybe if in doubt we can do as s.el does

vemv: (bump)

bbatsov: I can live with this, it just seems quite verbose.

.circleci/config.yml
```diff
@@ -9,11 +9,13 @@ default: &default-steps
     - run: make elpa
     - run: emacs --version
     - run: cask --version
+    - run: make unit
```
bbatsov, line 12: I think a target like `unit-tests` would be better here. `make test` should probably run all tests and there should be a separate `make` target for the integration tests.

tests/unit-test.el
```diff
@@ -0,0 +1,62 @@
+(require 'paredit)
+(require 'clj-refactor)
+
+(describe "completable-for-cljr-slash?"
+  (it "Returns `t' for tokens that can represent alias-prefixed named things (symbols, keywords, metadata)"
+    (expect (completable-for-cljr-slash? nil) :to-be nil)
+    (dolist (prefix '(""
```
expez, line 7: What about `#'my-alias/private-var`? 

Think that's the only other form that I use regularly that isn't covered.

clj-refactor.el
```diff
@@ -1934,14 +1934,19 @@ FEATURE is either :clj or :cljs."
...
        (string-equal "js" alias)))
 
+(defun cljr--namespace-of-thing-at-point ()
+  (thread-last (buffer-substring-no-properties
+                (cljr--point-after 'paredit-backward)
+                (1- (point)))
+    (string-remove-prefix "^::")
+    (string-remove-prefix "::")
+    (string-remove-prefix "^")
+    (string-remove-prefix "@")))
```
expez, line 1944: `#'`?

tests/unit-test.el
vemv: Good call! backtick also 

(unrelated, I think some CIDER features don't parse #' or backtick either...)

bbatsov:
Seems we forgot about this one.

vemv:
Will give it some Elisp love soon enough :)

PR #495, opened 2021-09-30
Make the refactor-nrepl version customizable

vemv:
Allows users to try out newer refactor-nrepl releases, which shouldn't be necessarily coupled to clj-refactor.el.


clj-refactor.el
```diff
@@ -3254,6 +3254,17 @@ if REMOVE-PACKAGE_VERSION is t get rid of the (package: 20150828.1048) suffix."
         (replace-regexp-in-string " (.*)" "" version)
       version)))
 
+(defcustom get-cljr-version (lambda ()
```
bbatsov, line 3257: Why don't we make this a string that just defaults to the result of `cljr--version`? Also a `defcustom`'s name should start with `cljr`, so I'd name this `cljr-middleware-version` or something along those lines.

clj-refactor.el
```diff
@@ -3254,6 +3254,17 @@ if REMOVE-PACKAGE_VERSION is t get rid of the (package: 20150828.1048) suffix."
...
+                              (cljr--version t))
+  "A no-args function that should return the refactor-nrepl version to be used.
+
+You can customize this in order to try out new releases.
+
+If customizing it, you most likely should `(setq cljr-suppress-middleware-warnings nil)' too,
+for avoiding a warning that would be irrelevant for this case."
+  :group 'cljr
+  :type 'function)
```
bbatsov, line 3266: I'd also add a `:package-version` property here, so it's easy to spot when this was added.

clj-refactor.el
bbatsov: Or rather `cljr-injected-middleware-version`. Not sure how I named this in CIDER, but it'd be nice for us to be consistent across the packages.

vemv:
All feedback addressed. Cider's most similar variable is https://github.com/clojure-emacs/cider/blob/adfc1c940d452ace7d1dbea543024d422f60364c/cider.el#L409 . 

`required` would be confusing for something customizable so `injected` seems clearer.

clj-refactor.el
```diff
@@ -3254,6 +3254,17 @@ if REMOVE-PACKAGE_VERSION is t get rid of the (package: 20150828.1048) suffix."
...
+(defcustom cljr-injected-middleware-version (cljr--version t)
+  "The refactor-nrepl version to be injected.
+
+You can customize this in order to try out new releases.
+
+If customizing it, you most likely should `(setq cljr-suppress-middleware-warnings nil)' too,
+for avoiding a warning that would be irrelevant for this case."
+  :group 'cljr
+  :type 'string
+  :package-version "3.0.0")
```
vemv, line 3266: I'm assuming next release will be 3.0.0 btw, for matching the refactor-nrepl version we intend to cut.

bbatsov (approved):
Looks good to me now. The test failures seem unrelated.

bbatsov:
> required would be confusing for something customizable so injected seems clearer.

Yeah, the name there is a bit confusing as it's also both the required version (you'll get a warning if your `cider-nrepl` version is older) and the injected version. Probably I should have used 2 separate defcustoms.

clj-refactor.el
bbatsov: Yep, I assume so as well. At some point it'd be nice to drop the lockstep in the version, something CIDER did a few years ago. (mostly for the sake of other consumers of the middleware)

vemv:
Cheers for the prompt review 🍻

bbatsov:
You're welcome!

clj-refactor.el
vemv, line 3266: I realised cider.el has `:safe #'<pred>` for all its defcustoms, should I do the same here? Does it make a difference? @bbatsov

bbatsov: It's not required but it's useful. Otherwise you'd get warnings when setting something via dir locals.

PR #497, opened 2021-09-30
Partially revert "[Fix #496] Drop support for Emacs 25"

### LINKED ISSUE #496: Drop support for Emacs 25
Basically CIDER had to do this, as parseedn relies now on `map.el` which requires Emacs 26. Not a big deal, as I expect almost no one uses Emacs 25 at this point. See https://github.com/clojure-emacs/cider/pull/3060 for more details.

vemv:
Partially reverts commit ecad41a5881111d620fe5735f95a28a003a7f8aa.

It seems fine to me to remove it from CI but I don't think we should drop compat if a presumably non-problematic hack can support it.

For one thing I happen to use 25 atm (I used 26 earlier but with Big Sur something changed). Sadly there are breaking changes across Emacs versions (25 -> 26 -> 27), some of them _very_ hard to track for non- Emacs insiders like me, so using an old Emacs isn't entirely out of the question.


vemv:
Checking out https://github.com/clojure-emacs/cider/pull/3060 the rationale doesn't seem superfluous. I know that sometimes complexity comes from upstream dependendencies (Emacs itself, map.el) so it's relatively little what we can do about that -__-

PR #498, opened 2021-10-03
Use the latest middleware

vemv:
Proposes using the latest middleware (the 3.* series is much more stable, no breaking changes) for clj-refactor.el users using the latest snapshot.

This way we can get extra user testing.

After releasing the 3.0.0 middleware (no alpha), we can undo this change and also bump clj-refactor.el itself to 3.0.0.

See also https://github.com/clojure-emacs/refactor-nrepl/pull/337


expez (approved):
This sounds like a good idea to me 👍

vemv:
Cheers 🍻

Issue #139, opened 2015-04-25
clean-project wants to save the world


philjackson:
Running `cljr-project-clean` seems to prompt continuously for a save file location (without it being clear what file needs to be written).

On latest release in Melpa (20150421.529).

magnars:
Nice issue title!

I fixed this problem in https://github.com/clojure-emacs/clj-refactor.el/commit/da2d02d030adaf6dab0e59f925d0490574ccd49a

benedekfazekas:
thx @magnars!

philjackson:
:thumbsup:

ccann:
I have this issue with:

clj-refactor 20170126.118
clojure-mode 5.6.1
cider 0.14.0
refactor-nrpl 2.3.0-SNAPSHOT
tools.nrepl 0.2.12
cider-nrepl 0.14.0

it just prompts me for a place to save some mystery file without specifying which file is being saved or why

vemv:
Same!

Using:

cider-20170525.255
clj-refactor-20170520.323
clojure-mode-20170527.1442

vemv:
Any clue? I can reproduce the issue with a pristine spacemacs, for example.

yogsototh:
Same issue with current `doom-emacs`.

vemv:
It looks like a bad interaction with undo-tree: https://grep.app/search?q=File%20to%20save%20in&filter[lang][0]=Emacs%20Lisp

vemv:
Looks like it would be a matter of temporarily removing `'undo-tree-save-history-from-hook` / `'undo-tree-save-history-hook` from `write-file-functions`. (The name of the hook will depend on your undo-tree version)

Seems simple enough but it someone wants to be a hero and QA this himself + PR it, so much the better 🍻 personally I use this refactor-nrepl feature  from a repl, not from clj-refactor.el so I have it slightly more difficult to try this out.

vemv:
Fix available in 3.1.0

PR #503, opened 2021-11-04
Prefer `find-file-noselect` to `with-temp-file`

vemv:
`with-temp-file` causes `(buffer-file-name)` to return nil which is the root cause of #139.

Additionally, now `cljr--update-file` does not leave files open (unless they were alrady open), improving performance / UX.

Fixes https://github.com/clojure-emacs/clj-refactor.el/issues/139

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


vemv:
This worked nicely in my machine.

Anyone else has the chance to try it out / review? @expez maybe?

expez (approved):
LGTM 👍

vemv:
Cheers ☕️☕️

PR #522, opened 2022-07-10
Refactor to support magic namespace require with language context annotation

dgtized:
Per my comment in https://github.com/clojure-emacs/clj-refactor.el/pull/521#issuecomment-1179577092, this PR makes no changes in the current behavior, but extracts and breaks up the current logic to facilitate using a defcustom in a later PR to select the magic require logic. In addition, it adds test coverage for each the methods extracted. My intention is to rebuild the changes in #521, as a sequence of commits on top of this. My hope is if this is merged, we can introduce a defcustom feature flag inside of cljr-slash, that will re-use some of these helpers but will otherwise be an independent change.

For review, I would recommend going through the first 3 commits step by step as they are clean changes, everything after I can squash as a single commit if preferred as they are just adding test coverage to the extracted methods.

Thanks in advance for your review!

---

Thanks!


vemv:
Thanks much for the splitting! I'll review tomorrow-ish

tests/unit-test.el
```diff
@@ -9,3 +9,67 @@
   (it "returns the ns name of its argument"
     (expect (cljr--ns-name "com.corp.foo") :to-equal "foo")
     (expect (cljr--ns-name "foo") :to-equal "foo")))
+
+(defmacro with-clojure-file (filename &rest body)
```
vemv, line 13:
```suggestion
(defmacro cljr--with-clojure-temp-file (filename &rest body)
```

clj-refactor.el
```diff
@@ -1963,29 +1963,35 @@ following this convention: https://stuartsierra.com/2015/05/10/clojure-namespace
         (gethash :clj aliases)
       (gethash :cljs aliases))))
 
+;; FIXME: this identifies "js" as a namespace in a cljc file, which is valid for
+;; clj context, but perhaps should not if alias is inside a #?(:cljs) context.
```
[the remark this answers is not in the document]
vemv: I'd be most comfortable with an issue

You've done an excellent writeup already so the issue can be a link to this comment simply - you call!

tests/unit-test.el
```diff
@@ -9,3 +9,67 @@
...
+
+  (it "returns js for js alias in clj file."
+    (with-clojure-file "foo.clj"
+      (insert "(ns foo)")
+      (expect (cljr--unresolved-alias-ref "js") :to-equal "js")))
+
+  (it "returns js for js alias in cljc file."
+    (with-clojure-file "foo.cljc"
+      (insert "(ns foo)")
+      (expect (cljr--unresolved-alias-ref "js") :to-equal "js"))))
```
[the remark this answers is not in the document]
dgtized: I think I was modeling this on https://github.com/clojure-emacs/clojure-mode/blob/master/test/clojure-mode-font-lock-test.el#L35, https://github.com/clojure-emacs/cider/blob/master/test/cider-font-lock-tests.el#L32, https://github.com/clojure-emacs/cider/blob/master/test/cider-overlay-tests.el#L25, and https://github.com/clojure-emacs/cider/blob/master/test/cider-util-tests.el#L34, which generally seem to wrap the macro outside of the expect. I'm happy to follow your suggestion, but that's why I made that style assumption. 

I was seeing test failures locally inside of the macro during development but I could see how it's possible it could be swallowed in the macro. I guess if it's outside it also disallows any expectations based on the current buffer state? Ie you couldn't assert against the value of `(point)` or the current active region outside of the context of the temp buffer in the expected value. That said, that might not be the best way to make those assertions anyway.

Happy to change it if that's your style preference, just wanted to give the context.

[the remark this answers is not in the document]
dgtized: Oh, I just went to fix this and I noticed https://github.com/clojure-emacs/clj-refactor.el/pull/522/files#diff-f78a6cdcd522a7fae3dbadcc142377d69570a8088e0fcb28c4f6d51ae0f7b8fcR22-R32. If the expect needs to be on the outside I guess that means I need to split the test into two expectations one for each step. I feel it's clearer with multiple examples to keep context outside (and avoids a bunch of test setup duplication), but can also appreciate if you would prefer to have one expectation per test. Any preferences here?

clj-refactor.el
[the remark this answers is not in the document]
dgtized: Thanks, added issue https://github.com/clojure-emacs/clj-refactor.el/issues/523.

tests/unit-test.el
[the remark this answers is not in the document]
vemv: I'd like sth like the following:

```lisp
(describe "cljr--insert-require-libspec"
  (let* ((after-inserting-a)
         (after-inserting-b))
    (with-clojure-file "libspec.clj"
      (insert "(ns lib)\n")
      (cljr--insert-require-libspec "[com.a :as a]")
      (setq after-inserting-a (buffer-string))
      (cljr--insert-require-libspec "[com.b :as b]")
      (setq after-inserting-b (buffer-string))))
  (it "adds a new require, over an empty ns form"
    (expect after-inserting-a :to-equal "..."))
  (it "adds a new require, over a ns form with existing requires"
    (expect after-inserting-b :to-equal "...")))
```

tests/unit-test.el
dgtized: fixed

dgtized:
I believe I addressed all concerns in the latest push. Let me know if anything else needs to be addressed.

tests/unit-test.el
```diff
@@ -9,3 +10,81 @@
...
+              (cljr--ns-alias-at-point))
+            :to-equal "aba"))
+
+  (it "removes namespace keyword ::"
+    (expect (with-temp-buffer
+              (insert "::ns-name/")
+              (cljr--ns-alias-at-point))
+            :to-equal "ns-name"))
+
+  (it "removes deref operator"
```
vemv, line 55: Couple more things that shoud be likely removed:

```
`
#'
```

vemv (commented):
Almost there

tests/unit-test.el
dgtized: Oh, I think that's new behavior. I'm happy to add more cases, but was trying to keep this focused on the current behavior.

I also noticed it didn't seem to like dotted namespace aliases like `clojure.set/` as it just returned "set". So I think there are a couple of cases we should account for. My preference would be to make that a different issue, which I'm also happy to tackle but trying to keep new behavior separate. What is your preference here?

vemv: Alright, let's create an issue and briefly link to it in a fixme

dgtized: https://github.com/clojure-emacs/clj-refactor.el/issues/524

dgtized: Added fixme + pending tests with https://github.com/clojure-emacs/clj-refactor.el/pull/522/commits/ead4813737ea4e21d5aa1c8422a1ba247321b481

vemv:
Excellent!

PR #527, opened 2022-07-17
Fix regex for removing prefix from ns alias

dgtized:
Fixes failing test in https://github.com/clojure-emacs/clj-refactor.el/pull/526.

The leading ^ matches the beginning of the line, and not from the current point.
Many of the examples were inserted as a bare word and not tested in the context
they would usually appear.

Also removed unnecessary escaping of character class [] brackets.

I've run buttercup unit-tests and the feature specs locally. However I was unable to get CircleCI to run integration tests for this branch due to the following environment issue:
```
cask exec ecukes --no-win
Please set the environment variable TERM; see 'tset'.
make: *** [Makefile:32: integration-tests] Error 1
```

It did however, pass unit tests. Is there an additional setting in CircleCI to fix that I need to set in my configuration?

Thanks!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,6 +2,8 @@ version: 2.1
 
 # Default actions to perform on each Emacs version
 default: &default-steps
+  environment:
+    TERM: xterm
   steps:
     - checkout
     - run: apt-get update && apt-get install make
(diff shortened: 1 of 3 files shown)


dgtized:
I see in the build failure environment setup for https://app.circleci.com/pipelines/github/clojure-emacs/clj-refactor.el/124/workflows/62577b8f-fc8f-4a27-8f0f-384462f983fd/jobs/378 that it reported:

```
Using environment variables from project settings and/or contexts:
  TERM=**REDACTED**
```

Is that actually a secret value or should that be specified in the config using https://circleci.com/docs/env-vars?

dgtized:
I amended the circle-ci config to include `TERM=xterm` so that it runs without manually specifying that in the environment. I can back that out if that's a problem but I think that is an easier workflow for development (provided there is no custom expectation for the terminal to execute in).

vemv:
Looks great. Thanks much for the detailed explanation!

vemv:
It's green now ✅

Issue #531, opened 2022-08-20
Create cljr-slash variant that uses the `suggest-libspecs` middleware op


vemv:
Create a cljr-slash variant that uses the `suggest-libspecs middleware` (https://github.com/clojure-emacs/refactor-nrepl/issues/384), and has a more streamlined UI than current cljr-slash (since it would never ask for language context).

These are the inputs it has to send to the middleware:

* The user input, e.g. `set` when the user typed `set/`
* The filename extension, grabbed from `(buffer-file-name)`
* The entire value of `cljr-magic-require-namespaces`

It will receive the following response, order by priority:

```
["[namespace :as alias]"
 "[another-namespace :as alias]"]
```

All that has to be done Emacs side is letting the user choose one of those, and insert it.

No logic should be performed here; that is done refactor-nrepl side.

The UI for letting the user choose should be customizable, in an Emacs-idiomatic fashion.

The variant should be opt-in, and disabled by default.

vemv:
@dgtized: the latest clj-refactor (3.5.6; will be available within a couple hours) / refactor-nrepl (3.5.5; released) include a draft of the new middleware, namely:

https://github.com/clojure-emacs/refactor-nrepl/commit/8825e26989726a849f0910988c764f136f85e8fc

which you can try like this:

```lisp
(thread-first "cljr-suggest-libspecs"
  (cljr--create-msg "lib-prefix" "s"
                    "language-context" "cljc"
                    "preferred-aliases" (prin1-to-string cljr-magic-require-namespaces))
  (cljr--call-middleware-sync "suggestions")
  parseedn-read-str)
```

as you can check out in the commit, this draft is merely focused in getting the API right (input and output format). The strings to be produced are subject to (much) improvement. For now we just return essentially the same as the old op `namespace-aliases`, with some basic context (language) awareness.

I hope this helps getting the ball rolling.

dgtized:
Started working on this, and I have a decent implementation, but went back to add tests and realized the build is broken from your last two commits. Do you mind addressing that?

vemv:
Hi! Excited to hear that.

The build is failing for a random reason, back then I did verify that even re-triggering a build from the latest green commit now would fail.

Obviously it has to be addressed, but you can rest assured that nothing is broken.

dgtized:
[...]

Issue #530, opened 2022-08-02
Ability to specify context in `cljr-magic-require-namespaces`


vemv:
`cljr-magic-require-namespaces` doesn't currently allow specifying its intended context (:clj, :cljs).

This can easily cause incorrect suggestions for cljr-slash.

I'd suggest one of the following:

* Introduce new defcustom, e.g. `cljr-jvm-magic-require-namespaces`
  * This would be a list like `'("clojure.java.io")`, specifying which values of `cljr-magic-require-namespaces` are intended for JVM clojure.
  * Same for cljs
* Allow cljr-magic-require-namespaces to also include lists
  * Example difference:

```diff
(defcustom cljr-magic-require-namespaces
  '(("set"  . "clojure.set")
-    ("io"   . "clojure.java.io")
+    ("io"    "clojure.java.io" "clj")
+    ("io"    "foo.io" "cljs")
    ("str"  . "clojure.string")
    ("walk" . "clojure.walk")
    ("zip"  . "clojure.zip")))
```

i.e. we could mix and match cons cells (`("set"  . "clojure.set")`) with lists having the extra property `("io"    "clojure.java.io" "clj")`.

Both approaches are intended to be backwards-compatible.

dgtized:
I'm curious why the defcustom should still be an alist prefixed by the namespace refer. If the intention is to send the entire defcustom to the middleware, can't we parse the recommended as target in the middleware? I wonder if we could make it a list of the recommendation string with a list of valid language contexts, where a nil list indicates all language contexts are allowed.

```
(defcustom cljr-magic-require-namespaces
  '(("clojure.set :as set")
    ("clojure.java.io :as io" ("clj"))
    ("foo.io :as io" ("cljs"))
    ("clojure.string :as str")
    ("clojure.walk :as walk")
    ("clojure.zip :as zip")))
```

In the future this could support examples where the recommended namespace contained a refers or a reader conditional. Ie an example like: `'("clojure.set :refer [union] :as set")`. I could also imagine prefixing the restrictions with a keyword like `:include`/`:exclude`, or `:only`/`:except` to account for the other language contexts possibilities.  Ie something like: `("clojure.java.io :as io" :only ("clj"))` or `("foo.io :as io" :except ("clj"))`.

[...]

PR #532, opened 2022-09-04
`cljr-slash` using `suggest-libspecs` middleware op

dgtized:
Implements https://github.com/clojure-emacs/clj-refactor.el/issues/531. ~~Still need to add changelog/readme updates if this approach is accepted.~~

If the defcustom `cljr-slash-uses-suggest-libspec` is enabled, `cljr-slash` will query the new `suggest-libspec` middleware op for suggested namespaces.

Currently it sends over the entire `cljr-magic-require-namespaces` as the preferred-aliases map, that may require pre-processing at some point, but the placeholder seems correct. It also correctly identifies if code block is in a conditional language context, and provides that information to the middleware to in order to assist generating recommendations.

Finally, as it's using `completing-read` to select namespaces, if the user inserts their own content into the response, that will be inserted as the new libspec. This allows the user to reference a new libspec, or edit it to include refers or require-macro. In the future, it might make sense to use the universal argument to temporarily force `cljr-magic-requires` to use :prompt, allowing a user to insert a require of their own choice more easily.

- [x] Fix wiki docs to reference the `cljr-slash-uses-suggest-libspec` defcustom

Thanks!

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -2134,23 +2134,26 @@ will add the corresponding require statement to the ns form."
                             (clojure-find-ns)
                             (cljr--unresolved-alias-ref (cljr--ns-alias-at-point))))
     (if cljr-slash-uses-suggest-libspec
+        ;; New path creates suggestions from `suggest-libspec' middleware op
         (when-let (libspec
                    (thread-first
                      alias-ref
                      (cljr--call-middleware-suggest-libspec (cljr--language-context-at-point))
                      cljr--prompt-or-select-libspec))
+          ;; only insert a require if a candidate exists and was selected
           (cljr--insert-require-libspec libspec))
-        (when-let (aliases (cljr--magic-requires-lookup-alias alias-ref))
-          (let ((short (cl-first aliases))
-                ;; Ensure it's a list (and not a vector):
-                (candidates (mapcar 'identity (cl-second aliases))))
-            (when-let (long (cljr--prompt-user-for "Require " candidates))
-              (when (and (not (cljr--in-namespace-declaration-p (concat ":as " short "\b")))
-                         (not (cljr--in-namespace-declaration-p (concat ":as-alias " short "\b")))
-                         (or (not (eq :prompt cljr-magic-requires))
-                             (not (> (length candidates) 1)) ; already prompted
-                             (yes-or-no-p (format "Add %s :as %s to requires?" long short))))
-                (cljr--insert-require-libspec (format "[%s :as %s]" long short)))))))))
+      ;; Old path creates suggestions from `namespace-aliases' middleware op
+      (when-let (aliases (cljr--magic-requires-lookup-alias alias-ref))
+        (let ((short (cl-first aliases))
+              ;; Ensure it's a list (and not a vector):
+              (candidates (mapcar 'identity (cl-second aliases))))
+          (when-let (long (cljr--prompt-user-for "Require " candidates))
+            (when (and (not (cljr--in-namespace-declaration-p (concat ":as " short "\b")))
+                       (not (cljr--in-namespace-declaration-p (concat ":as-alias " short "\b")))
+                       (or (not (eq :prompt cljr-magic-requires))
+                           (not (> (length candidates) 1)) ; already prompted
+                           (yes-or-no-p (format "Add %s :as %s to requires?" long short))))
+              (cljr--insert-require-libspec (format "[%s :as %s]" long short)))))))))
 
 (defun cljr--in-namespace-declaration-p (s)
   (save-excursion
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -1965,13 +1965,19 @@ following this convention: https://stuartsierra.com/2015/05/10/clojure-namespace
     parseedn-read-str))
 
 (defun cljr--call-middleware-suggest-libspec (alias-ref language-context)
-  "Suggest libspec entries for an `alias-ref' in a `language-context'.
+  "Suggest namespace require libspecs for an alias.
 
-Returns a candidate list of libspec entry strings. `alias-ref' is
-a string representing a namespace alias. `language-context' is a
-2 element list representing the language context of the buffer,
-followed by the language context at the current point. Assume
-same context as buffer if context at current point is nil.
+Returns a candidate list of libspec entry strings. An alias
+\"set\" might return (\"[clojure.set :as set]\").
+
+`alias-ref' is a string representing a namespace alias entered by
+the user.
+
+`language-context' is a 2 element tuple representing the language
+context of the buffer, and the context of the reader conditional
+where the alias is being used. Assume same context as
+buffer if context at current point is nil. See
+`cljr--language-context-at-point' for details on this tuple.
 
 Passes through the custom `cljr-magic-require-namespaces' so that
 users can specify default recommended alias prefixes that may not
@@ -1994,7 +2000,14 @@ appear in the project yet."
 
 Returns a tuple, the first value represents the language context
 of the file, the second represents the language context at the
-current point if it is within a reader conditional. If a given value is unknown, it will be expressed as nil"
+current point if it is within a reader conditional. If a given
+value is unknown, it will be expressed as nil.
+
+Reader conditionals are forms like #?(:clj (expr)) or
+#?@(:cljs (expr) :default (expr2)) (see
+URL`https://clojure.org/guides/reader_conditionals'). As
+example, `(\"cljc\", \"cljs\")' represents a point in a cljs path
+of a reader conditional inside of a cljc file."
   (list (cond ((cljr--cljc-file-p)
                "cljc")
               ((cljr--cljs-file-p)
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -75,7 +75,7 @@ Any other non-nil value means to add the form without asking."
                  (const :tag "false" nil)))
 
 (defcustom cljr-slash-uses-suggest-libspec nil
-  "If t, `cljr-slash' magic require functionality will use the
+  "If t, `cljr-slash' magic require functionality will use the newer
 `cljr-suggest-libspec' middleware op instead of the
 `namespace-aliases' op."
   :type 'boolean
@@ -1994,8 +1994,7 @@ appear in the project yet."
 
 Returns a tuple, the first value represents the language context
 of the file, the second represents the language context at the
-current point if it is within a reader conditional. If either
-value is unknown, return nil."
+current point if it is within a reader conditional. If a given value is unknown, it will be expressed as nil"
   (list (cond ((cljr--cljc-file-p)
                "cljc")
               ((cljr--cljs-file-p)
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -1982,7 +1982,8 @@ buffer if context at current point is nil. See
 Passes through the custom `cljr-magic-require-namespaces' so that
 users can specify default recommended alias prefixes that may not
 appear in the project yet."
-  (seq-let (buffer-context point-context) language-context
+  (let ((buffer-context (car language-context))
+        (point-context (cadr language-context)))
     (thread-first
       "cljr-suggest-libspecs"
       cljr--ensure-op-supported
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -1957,6 +1957,31 @@ following this convention: https://stuartsierra.com/2015/05/10/clojure-namespace
     (cljr--call-middleware-sync "namespace-aliases")
     parseedn-read-str))
 
+(defun cljr--call-middleware-suggest-libspec (alias-ref language-context)
+  "Suggest libspec entries for an `alias-ref' in a `language-context'.
+
+Returns a candidate list of libspec entry strings. `alias-ref' is
+a string representing a namespace alias. `language-context' is a
+2 element list representing the language context of the buffer,
+followed by the language context at the current point. Assume
+same context as buffer if context at current point is nil.
+
+Passes through the custom `cljr-magic-require-namespaces' so that
+users can specify default recommended alias prefixes that may not
+appear in the project yet."
+  (seq-let (buffer-context point-context) language-context
+    (thread-first
+      "cljr-suggest-libspecs"
+      cljr--ensure-op-supported
+      (cljr--create-msg "lib-prefix" alias-ref
+                        "language-context" buffer-context
+                        "buffer-language-context" buffer-context
+                        "input-language-context" (or point-context buffer-context)
+                        "preferred-aliases" (prin1-to-string cljr-magic-require-namespaces))
+      (cljr--call-middleware-sync "suggestions")
+      parseedn-read-str
+      (seq-into 'list))))
+
 (defun cljr--get-aliases-from-middleware ()
   (when-let (aliases (cljr--call-middleware-for-namespace-aliases))
     (if (cljr--clj-context-p)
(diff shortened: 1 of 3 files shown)


clj-refactor.el
```diff
@@ -74,6 +74,13 @@ Any other non-nil value means to add the form without asking."
                  (const :tag "prompt" :prompt)
                  (const :tag "false" nil)))
 
+(defcustom cljr-slash-uses-suggest-libspec nil
+  "If t, `cljr-slash' magic require functionality will use the
```
vemv, line 78:
```suggestion
  "If t, `cljr-slash' magic require functionality will use the newer
```

clj-refactor.el
```diff
@@ -1957,6 +1964,61 @@ following this convention: https://stuartsierra.com/2015/05/10/clojure-namespace
     (cljr--call-middleware-sync "namespace-aliases")
     parseedn-read-str))
 
+(defun cljr--call-middleware-suggest-libspec (alias-ref language-context)
+  "Suggest libspec entries for an `alias-ref' in a `language-context'.
```
vemv, line 1968: Would seem nice to explain what both things are

clj-refactor.el
```diff
@@ -1957,6 +1964,61 @@ following this convention: https://stuartsierra.com/2015/05/10/clojure-namespace
...
+Returns a candidate list of libspec entry strings. `alias-ref' is
+a string representing a namespace alias. `language-context' is a
+2 element list representing the language context of the buffer,
+followed by the language context at the current point. Assume
+same context as buffer if context at current point is nil.
+
+Passes through the custom `cljr-magic-require-namespaces' so that
+users can specify default recommended alias prefixes that may not
+appear in the project yet."
+  (seq-let (buffer-context point-context) language-context
```
vemv, line 1979: `seq-let` is probably nice but this lib (and I, coincidentally) only use seq.el for "seq stuff", not for other conveniences which would increase cognitive load to casual maintainers.

I'm generally happy with `let*` (and prefer `let*` over `let`, no matter what)

clj-refactor.el
```diff
@@ -2059,17 +2121,24 @@ will add the corresponding require statement to the ns form."
...
-            (candidates (mapcar 'identity (cl-second aliases))))
-        (when-let (long (cljr--prompt-user-for "Require " candidates))
-          (when (and (not (cljr--in-namespace-declaration-p (concat ":as " short "\b")))
-                     (not (cljr--in-namespace-declaration-p (concat ":as-alias " short "\b")))
-                     (or (not (eq :prompt cljr-magic-requires))
-                         (not (> (length candidates) 1)) ; already prompted
-                         (yes-or-no-p (format "Add %s :as %s to requires?" long short))))
-            (cljr--insert-require-libspec (format "[%s :as %s]" long short))))))))
+    (if cljr-slash-uses-suggest-libspec
+        (when-let (libspec
```
vemv, line 2125: Maybe add a `;; comment` explaining the case where `libspec` is nil

clj-refactor.el
```diff
@@ -1957,6 +1964,61 @@ following this convention: https://stuartsierra.com/2015/05/10/clojure-namespace
...
+      parseedn-read-str
+      (seq-into 'list))))
+
+(defun cljr--language-context-at-point ()
+  "Detects the current language-context at a given point.
+
+Returns a tuple, the first value represents the language context
+of the file, the second represents the language context at the
+current point if it is within a reader conditional. If either
+value is unknown, return nil."
```
vemv, line 1998:
```suggestion
current point if it is within a reader conditional. If a given value is unknown, it will be expressed as nil"
```

or something to that effect. e.g. we never return nil, we always return a tuple

tests/unit-test.el
```diff
@@ -141,31 +141,132 @@
...
+    (cljr--with-clojure-temp-file "foo.cljc"
+      (insert "(ns foo)")
+      (expect (cljr--unresolved-alias-ref "js")
+              :to-equal "js"))))
+
+(defmacro with-point-at (text &rest body)
+  (declare (indent 1))
+  `(progn (insert ,text)
+          (goto-char (point-min))
+          (re-search-forward "|")
```
vemv, line 177: Let's use `<POINT>` as the marker, which although ugly, is unmistakable and will result in pretty self-describing tests IMO

tests/unit-test.el
```diff
@@ -141,31 +141,132 @@
...
+
+  (it "identifies a cljc file"
+    (cljr--with-clojure-temp-file "foo.cljc"
+      (insert "(ns foo)")
+      (expect (cljr--language-context-at-point)
+              :to-equal '("cljc" nil))))
+
+  (it "identifies a cljc file with a cljs context"
+    (cljr--with-clojure-temp-file "foo.cljc"
+      (with-point-at "(ns foo) #?(:cljs (Math/log|))"
```
vemv, line 202: We could exercise a few more cases:

* `?(:clj 1 :cljs (Math/log|)`
  * two branches present
* `?(:cljs (Math/log| :clj 1)`
  * two branches present, reverse order
* `?(|)`
  * incomplete reader conditional

tests/unit-test.el
```diff
@@ -141,31 +141,132 @@
...
+
+  (it "returns user content from prompt regardless of candidates"
+    (spy-on 'completing-read :and-return-value "[alpha :as a]")
+    (expect (cljr--prompt-or-select-libspec
+             '("[a.a :as a]"
+               "[a.b :as b]"))
+            :to-equal "[alpha :as a]"))
+
+  (it "short-circuits if only one candidate matches"
+    (expect 'completing-read :to-have-been-called-times 0)
```
vemv, line 227: Is it possible to add a counterexample where completing-read is called `> 0` times?

That would make this test more logically robust.

tests/unit-test.el
```diff
@@ -141,31 +141,132 @@
...
+
+  (it "prompts anyway if `cljr-magic-requires' is `:prompt'"
+    (spy-on 'completing-read :and-return-value "[a.a :as a]")
+    (let ((cljr-magic-requires :prompt))
+      (expect (cljr--prompt-or-select-libspec '("[a.a :as a]"))
+              :to-equal "[a.a :as a]"))))
+
+(describe "cljr-slash"
+  (before-each (setq cljr-slash-uses-suggest-libspec t))
+  (after-each (setq cljr-slash-uses-suggest-libspec nil))
```
vemv, line 239: Let's set it to `<whatever the previous value was>` instead of `nil`

tests/unit-test.el
```diff
@@ -141,31 +141,132 @@
...
+    (spy-on 'cljr--call-middleware-suggest-libspec
+            :and-return-value (parseedn-read-str "[\"[bar.alias :as alias]\"]"))
+    (cljr--with-clojure-temp-file "foo.cljc"
+      (with-point-at "(ns foo)\nalias|"
+        (cljr-slash))
+      (expect (buffer-string) :to-equal "(ns foo
+  (:require [bar.alias :as alias]))
+alias/")))
+
+  (it "prompts user for selection from suggest-libspec"
```
vemv, line 250: Fantastic tests, thank you 👏

vemv (approved):
This looks great to me. Thanks much!

I assume you've tried this over the real middleware we released for refactor-nrepl.

* Does it work correctly?
* Do you need this PR merged asap, or are you comfortable by working with your clj-refactor fork day-to-day?
* Do you have any new feedback item for the refactor-nrepl middleware, or shall we implement it as planned?

Cheers - V

clj-refactor.el
dgtized: Both `alias-ref` and `language-context` are described in the next paragraph. If you have suggestions on how to improve those descriptions for clarity I am happy to change them, but I need more details on the concern here.

clj-refactor.el
dgtized: I think this will be easier to follow once the `cljr-slash-uses-suggest-libspec` feature flag is gone. If we can't find a libspec we don't suggest or insert one, which is the same behavior we used previously but required two `when-let`. We could extract the case where `cljr-slash-uses-suggest-libspec` is nil into it's own function, maybe that would be clearer? Could you elaborate on what your concern is here, I don't think I understand.

tests/unit-test.el
dgtized: That's what the other three tests are doing that are replacing the return value for 'completing-read with a `spy-on` with a `:and-return-value` mock return value. The reason this particular example is asserting it's not called is because it's one of the few cases it's not in the call graph. There may be a counter-example I missed though, so I would be happy to add one if I misunderstood, but I need more explanation.

tests/unit-test.el
dgtized: I don't think the current method of parsing will handle all of that. I cribbed something together to support a few cases using the existing reader conditional code. However, this should also support, :cljr, :bb, and :default and the incomplete reader conditional, and all of those cause emacs to lockup in an infinite loop as the existing `cljr--point-in-reader-conditional-branch-p` method expects an exact match or loops forever. I'm not sure how critical this is just yet though, as from my understanding the middleware is still ignoring this field.

I've add the test cases here and mark the ones that are failing as pending but I think it may out of scope to address this in this PR as it's fixes for existing code. This should definitely be addressed before the flag is enabled by default though.

tests/unit-test.el
dgtized: This is behavior is broken in buttercup anyway. It's supposed to support using :var ((cljr-slash-uses-suggest-libspec t)) as a lexical let around the entire describe but it just doesn't appear to work at present. I just swapped it with a local lexical let for each cljr-slash invocation to enable the feature for each call.

tests/unit-test.el
dgtized: I cribbed this approach from `clojure-mode`, `cider`, and `smartparens` tests. I would much prefer to match a common formatting scheme over inventing our own. You are welcome to change how you see fit though.

vemv: > I cribbed this approach from clojure-mode, cider, and smartparens tests.

Ok, let's go with it.

tests/unit-test.el
vemv: > and all of those cause emacs to lockup in an infinite loop

Well, that seems an important issue, showing that it was a good use of code review. We most certainly don't want people experiencing/reporting infinite loops.

(Especially since this PR will use that faulty code more frequently than we do now)

It would be a much welcome side quest if you could fix this (here or separately - as you wish)

tests/unit-test.el
vemv: If, as a maintainer, I read `(expect 'completing-read :to-have-been-called-times 0)`, I will expect a counterexample which says `(expect 'completing-read :to-have-been-called-times <a positive number>)`, using the same construct, in a nearby test case.

Otherwise it is overassuming to expect people to detailedly understand all tests cases, and how test cases that use different constructs relate to each other.

Basically I'm suggesting to turn an implicit similarity into an explicit one.

tests/unit-test.el
vemv: Could you paste something similar as a `;; comment`?

Linking to a buttercup issue or such

dgtized: I don't have a buttercup issue for it, it just didn't work. I just changed it to use:

```
(let ((cljr-slash-uses-suggest-libspec t))
          (cljr-slash))
```

For those three test cases. That's correct though a little more verbose. If someone wants to clean it up or understands buttercup better so they can get it to work they are welcome to, I don't have the time to tweak it though, sorry.

tests/unit-test.el
dgtized: Turns out I misread how the count expectations work, it has to be the at the end of the test. I thought it worked like rspec where it asserts at the beginning and then checks all assertions at the end. I added call count assertions at the end of all 4 of those test cases.

clj-refactor.el
dgtized: The [emacs docs style guide](https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html) recommends 67 characters for the first line of a documentation so it looks better in apropos. . I was attempting to keep the documentation intro simple and explain it in more detail later to match that style recommendation.

I'm going to change it to "Suggest namespace require libspecs for an alias", and then expand the documentation below to include examples, crosslinking to `cljr--language-at-point` and the corresponding clojure docs. That way the complexity of the variables description is not introduced until the paragraph below. I hope that resolves your concern.

clj-refactor.el
dgtized: Couldn't think of a name to keep this clear, so just left a comment. Formatting was off anyway.

tests/unit-test.el
dgtized: thanks!

tests/unit-test.el
dgtized: I'm going to temporarily disable the reader conditional language context field while it's locking up and disable the corresponding tests for now. That way it's not introducing a higher risk of causing emacs to block, and we can re-enable it when the middleware starts supporting that field and implement a more robust parsing solution using the existing tests. I think the API is fine here though, we can improve it as we move forward.

clj-refactor.el
dgtized: I've addressed this to your recommendation. From my perspective, destructuring binding is much easier to follow over manually parsing out car/cadr of each element, but to each their own. I appreciate your concern about cognitive load from unfamiliar constructs, but it's also frustrating and discouraging as a contributor to so frequently encounter code included in emacs core that is then dismissed at review. Do you have a style guide somewhere that I could follow to discover which are recommended and which are discouraged instead of discovering all of these at review time?

tests/unit-test.el
dgtized: Extracted into https://github.com/clojure-emacs/clj-refactor.el/issues/533

dgtized:
I believe I've addressed all outstanding concerns, updated external documentation and added a changelog entry as well as manually retesting it in a project. I would much prefer we merge this over letting it bitrot in review. It's not perfect but it will allow folks to start testing and working off of the new middleware so we can iterate from there. I believe the next steps beyond this PR on the cljr-refactor side are to address the reader conditional parsing in #533, and then document the new syntax for `cljr-magic-require-namespaces` in https://github.com/clojure-emacs/clj-refactor.el/issues/530, though I think all of the logic there is in the middleware, so it's probably better to document that after it's implemented.

I believe that API is solid enough to move forward with. I didn't encounter anything critical with it in development, but now that the logic has moved into the middleware, it's likely we won't see issues in the recommendations until we encounter them in actual use.

Let me know if there is any other critical concerns that need to be addressed prior to merge.

dgtized:
Just checking back in to see what else needs to be done here before it's merged.

vemv:
Thanks for the ping! Will check out today / tomorrow at most

clj-refactor.el
vemv: >  I could follow to discover which are recommended and which are discouraged instead of discovering all of these at review time?

Not that I know of

Generally, afaict in clojure-emacs .el code we lean toward simple constructs. Even then, different reviewers might have different preferences.

My genuine recommendation would be to simply be ready to tweak some things upon request. I myself cannot predict the feedback I'll receive in the variety of repos I contibute to - some little churn on superficial things is just part of life! :)

vemv:
Thanks again for this first step!

Should https://github.com/clojure-emacs/refactor-nrepl/issues/384 be implemented as planned? Anything that came to mind more recently?

dgtized:
I think we can proceed with the remainder of clojure-emacs/refactor-nrepl#384 as planned. However, until #533 is implemented, we won't know the correct current language context at point. My expectation remains that we will have more tweaking required on the exact recommendations from the backend, but that won't be as visible until it's in use. I'll see if I can take a stab at #533 in the coming week or two, though so it can operate with the correct local information.

Thanks for merging this, and the review!

vemv:
Nice! I'll try to advance on the middleware, do feel free to ping if you'd need to have it sooner.

PR #538, opened 2023-01-17
[Fix #537] Fail early in cljr-rename-symbol if narrowing is in effect

### LINKED ISSUE #537: `cljr-rename-symbol` not working properly when narrowing is in effect
## Expected behavior
When executing `cljr-rename-symbol` on a function name while [narrowing](https://www.gnu.org/software/emacs/manual/html_node/emacs/Narrowing.html) is in effect, `cljr-rename-symbol` should either execute the renaming as if no narrowing was in effect or issue a message about being unable to do so.   
I would of course prefer the first option, but I understand that this might be impossible, since the narrowing takes away context necessary for a thorough renaming in the whole file. Maybe a compromise would be to only execute the renaming in the narrowed buffer content? It would then be up to the user to ensure that the narrowed content contains all context relevant for the renaming.

## Actual behavior
`cljr-rename-symbol` claims (via a message in the echo area) to have renamed the function without having actually done that.

Side note: when trying to rename function arguments while narrowing is in effect, the message _cljr-rename-symbol: Wrong type argument: char-or-string-p, nil_ is issued (and the argument isn’t actually renamed).

1. Assuming that you have a running CIDER session, add the following function definition to a Clojure buffer:
```Clojure
(defn some-function [some-arg]
  (+ 1 2 3))
```
2. Narrow the buffer to the function definition by placing point within it and executing `narrow-to-defun`.
3. Place point on the function name, execute `cljr-rename-symbol`, answer _y_ on the first prompt, provide a new name on the second prompt and press _Enter_.
4. The message _Renamed 1 occurrences of some-function_ appears in the echo area, but the function name in the buffer hasn’t actually changed.
 
clj-refactor 3.6.0 (package: 20221023.1644), refactor-nrepl 3.6.0

### CIDER version information
CIDER 1.6.0 (Buenos Aires), nREPL 1.0.0  
Clojure 1.10.1, Java 11.0.8

### Leiningen version
2.10.0 on Java 11.0.8 OpenJDK 64-Bit Server VM

### Emacs version
GNU Emacs 27.1 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.24, cairo version 1.16.0) of 2021-03-28, modified by Debian

### Operating system
Linux 4.19.0-6-amd64 Debian 4.19.67-2+deb10u2 (2019-11-11) x86_64 GNU/Linux

rechvs:
Previously the function would report a successful renaming without actually having renamed anything.


rechvs:
I tried to add a test, but `make test` results in 

> cask exec buttercup -L .
Warning (emacs):
clj-refactor.el:78: First line is not a complete sentence
...
Failing due to checkdoc warnings...
make: *** [Makefile:35: unit-tests] Error 1

even when executed in the master branch, so I had no way of knowing whether my test would have actually worked. Is this expected? Or am I misinterpreting the output?  
(Taking the risk of stating the obvious here:) It would be nice if the CircleCI checks would work again, so I could compare my local setup with the CI one to find out what’s causing the tests to fail on my machine.

Nevertheless thanks a lot for your work on this project. :heart: After working with IntellIJ IDEA for quite some time now, I’ve really grown fond of its refactoring capabilities, so `cljr-rename-symbol` in particular is a real godsend.

clj-refactor.el
```diff
@@ -2826,6 +2826,7 @@ See: https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-find-usages"
 
 See: https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-rename-symbol"
   (interactive)
+  (if (buffer-narrowed-p) (error "Cannot rename symbols while narrowing is in effect"))
```
expez, line 2829: Think this needs to be `when` or the linter is going to complain about an `if` clause without corresponding `else` branch.

rechvs: Done.

bbatsov:
Won't it be better to just ignore the narrowing instead?

rechvs:
> Won't it be better to just ignore the narrowing instead?

As I said in #537: that would also be _my_ preferred option. But I’m of course in no position to decide whether that’s the way forward.

Issue #533, opened 2022-09-07
`cljr--point-in-reader-conditional-branch-p` may block and is incorrect


dgtized:
Extracted https://github.com/clojure-emacs/clj-refactor.el/pull/532#discussion_r965202125

In certain conditions `cljr--point-in-reader-conditional-branch-p` appears to block infinitely. It's not clear if it's in this method, or in `cljr--goto-reader-conditional` but the example tests in #532 introduce some test cases involving reader conditionals with `:default`, `:bb`, `:cljr` or missing a path that result in an infinite loop in the test.

I believe the issue here is this API is *asserting* we are on a specific path and returning nil if not. A better API would return nil if not in a reader conditional, or the language context of the reader conditional if we are. The revised method should handle all reader conditional cases for both #? and #?@.

```
(|) => nil
#? => nil
#?() => nil
#?(:clj (|)) => clj
#?(:cljs (|) :clj (expr)) => cljs
#?(:cljs (|) :clj (expr)) => cljs
#?(:cljs (|)) => cljs
#?(:cljs (expr) :clj (|)) => clj
#?(:bb (|) :cljs (expr)) => bb
#?(:cljr (|)) => cljr
#?(:default (|)) => default/nil?
#?(:default #?(:cljs (|)) => cljs?
```

It might make sense to nil pun default into nil since it's effectively falling upward into the top level context. I don't think I've ever seen a reader conditional nested inside another, but I think it's technically possible inside of a default block?

Whichever new method is introduced should replace the `(when (cljr--point-in-reader-conditional-p))` block of `cljr--language-context-in-point`, that was temporarily disabled.

dgtized:
Resolved now that both #541 and #542 are merged.

PR #542, opened 2023-04-28
Deprecate unused reader-conditional functions and transition to new

dgtized:
Cleanup PR to close out https://github.com/clojure-emacs/clj-refactor.el/issues/533. I initially marked each function as obsolete but then remembered these are conventionally private API so probably not worth the obsolete tag. We can re-apply obsolete if needed though by reverting the last commit. I also left TODOs on the remaining deprecated reader-conditional functions like `cljr--clj-context-p` to clarify they are being removed and when. Once we switch to enable `cljr-slash-uses-suggest-libspec`, there are quite a number of functions that can be removed, but I wanted to mark the ones specifically related to context in this PR.

One other thing, tests are passing on CI on my fork, but it appears tests are failing on the official repo because the circle CI ssh key for github is invalidated, and hasn't been refreshed? I suspect that requires a circle ci account owner for the main repo to resolve.

-- no need for tests on removed functions
-- I can add a changelog/readme if needed but it's not changing user-visible functionality.

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -234,6 +234,8 @@ won't run if there is a broken namespace in the project."
   :type 'boolean
   :safe #'booleanp)
 
+;; TODO: remove after `cljr--clj-context-p' is deprecated by enabling
+;; `cljr-slash-uses-suggest-libspec'.
 (defcustom cljr-assume-language-context nil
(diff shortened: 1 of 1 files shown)


dgtized:
Checking back in, did you get a chance to review? No worries if too busy, just making sure this doesn't fall off your radar.

PR #546, opened 2023-07-31
Increase `js/` namespace detection accuracy for .cljc files

### LINKED ISSUE #523: cljr-slash magic require js/ behavior is determined by file type, not language context
Context: https://github.com/clojure-emacs/clj-refactor.el/pull/522#discussion_r918170325

Specifically  this is discussing the behavior of `cljr--js-alias-p`, which ignores a magic require alias prefix "js" if and only if it's in a cljs file. It will attempt to lookup the namespace in a `cljc` file. It's less clear what the behavior is inside of a `#?(:clj)` or `#?(:cljs)` language context within a cljc file, but I would expect that it should use the cljs behavior inside of a cljs context in a cljc file.

Expected behavior is that `js/` in a cljs *context* should ignore the magic alias lookup for both cljs and cljc files.

vemv:
Fixes #523

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -1087,19 +1087,22 @@ If CLJS? is T we insert in the cljs part of the ns declaration."
 
 (defun cljr--cljs-file-p (&optional buf)
   "Is BUF, or the current buffer, visiting a cljs file?"
-  (string-equal (file-name-extension (buffer-file-name (or buf (current-buffer))))
-                "cljs"))
+  (when-let ((bfn (buffer-file-name (or buf (current-buffer)))))
+    (string-equal (file-name-extension bfn)
+                  "cljs")))
 
 (defun cljr--cljc-file-p (&optional buf)
   "Is BUF, or the current buffer, visiting a cljc file?"
-  (string-equal (file-name-extension (buffer-file-name (or buf (current-buffer))))
-                "cljc"))
+  (when-let ((bfn (buffer-file-name (or buf (current-buffer)))))
+    (string-equal (file-name-extension bfn)
+                  "cljc")))
 
 (defun cljr--clj-file-p (&optional buf)
   "Is BUF, or the current buffer, visiting a clj file?"
   (or (eq major-mode 'clojure-mode)
-      (string-equal (file-name-extension (buffer-file-name (or buf (current-buffer))))
-                    "clj")))
+      (when-let ((bfn (buffer-file-name (or buf (current-buffer)))))
+        (string-equal (file-name-extension bfn)
+                      "clj"))))
 
 (defun cljr--add-test-declarations ()
   (save-excursion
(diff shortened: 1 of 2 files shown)


clj-refactor.el
```diff
@@ -1087,19 +1087,22 @@ If CLJS? is T we insert in the cljs part of the ns declaration."
...
+  (when-let ((bfn (buffer-file-name (or buf (current-buffer)))))
+    (string-equal (file-name-extension bfn)
+                  "cljc")))
 
 (defun cljr--clj-file-p (&optional buf)
   "Is BUF, or the current buffer, visiting a clj file?"
   (or (eq major-mode 'clojure-mode)
-      (string-equal (file-name-extension (buffer-file-name (or buf (current-buffer))))
-                    "clj")))
+      (when-let ((bfn (buffer-file-name (or buf (current-buffer)))))
```
vemv, line 1103: Prevents nil access from e.g. IELM.

vemv (commented):
@dgtized: feel free to eye, although I understand you may be on lower availability.

clj-refactor.el
```diff
@@ -1087,19 +1087,22 @@ If CLJS? is T we insert in the cljs part of the ns declaration."
...
+        (string-equal (file-name-extension bfn)
+                      "clj"))))
```
dgtized, line 1105: Maybe we should extract this logic into a single `(cljr--file-type-p extension buffer)`? and re-use this helper for each of the cljc--file-p, clj-file-p, cljs-file-p?

Separately though, I don't think I noticed that major-mode `clojure-mode is sufficient for clj-file-p. That seems ... questionable if the file is cljs or cljc?

clj-refactor.el
```diff
@@ -2042,8 +2047,11 @@ is not set to `:prompt'."
       (gethash :cljs aliases))))
 
 (defun cljr--js-alias-p (alias)
-  (and (cljr--cljs-file-p)
-       (string-equal "js" alias)))
+  (and (string-equal "js" alias)
+       (member "cljs" (condition-case nil
```
dgtized, line 2051: I wonder if it would be clearer to only call member on the result of `(cljr--language-context-at-point)`, and just return t for the error/cljr--cljs-file-p case, instead of adding a possible list structure on the error branch to check with member if it falls through from the context-at-point case. Or is that problematic as the member call might trigger the error?

Alternatively, could we re-order this and do something like 

```
(and (string-equal "js" alias) 
         (or (cljr--cljs-file-p)
               (member (condition-case ... (error nil)))))
```
?

language-context-at-point *should* cover the case either way, but I wonder if shortcutting on the current cljs file first, and then falling into language-context-at point if we aren't certain it's cljs?

I may just be reacting to the potential error case in general though. Maybe we could wrap the body of cljr--language-context-at-point with a condition-case so it can never error and avoid handling edge cases in all of it's callers?

dgtized (commented):
Overall this makes sense, I guess a little cautious about some of the error handling but the logic makes sense to me. I didn't get a chance to load it up or verify tests though.

That said, the major-mode = 'clojure-mode check in a clj-file-p seems like it may causing other issues (particularly for anything that does a fall through checking clj-file-p first in a condition chain). I don't know if we should address that separately but I think we should stick to only checking file extensions there, and consider the major mode check for any callers where it's appropriate.

clj-refactor.el
vemv: CIDER uses this sort of logic:

```lisp
(cond
     ((derived-mode-p 'clojurescript-mode) ,,,)
     ((derived-mode-p 'clojurec-mode) ,,,)
     ((derived-mode-p 'clojure-mode) ,,,)
```

i.e. the clojure-mode mode is only to be expected for .clj files.

We could start using derived-mode-p though

PR #555, opened 2023-12-01
Expand cljr magic require namespaces documentation

dgtized:
* Minimal documentation on the :only keyword.
* Backport support for `:only` in the `cljr-magic-requires-namespace` if the suggest-libspec flag is off
* Adds edn alias and limits io alias to clj only

Leverages https://github.com/clojure-emacs/refactor-nrepl/pull/392 and https://github.com/clojure-emacs/clj-refactor.el/issues/530.

Still needs more testing and expanded documentation but this seemed like a good incremental improvement.

  - not adding test coverage here as this is the deprecated path, but I did verify the parser change at the elisp repl.

Thanks!

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -92,13 +92,16 @@ paths once this flag is removed."
 
 (defcustom cljr-magic-require-namespaces
   '(("edn"  . "clojure.edn")
-    ("io"   . "clojure.java.io")
+    ("io"     "clojure.java.io" :only ("clj"))
     ("math" . "clojure.math")
     ("set"  . "clojure.set")
     ("str"  . "clojure.string")
     ("walk" . "clojure.walk")
     ("zip"  . "clojure.zip"))
-  "Alist of aliases and namespaces used by `cljr-slash'."
+  "Alist of aliases to namespace libspec recommendations for `\\[cljr-slash]'.
+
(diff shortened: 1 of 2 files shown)


clj-refactor.el
```diff
@@ -91,13 +91,17 @@ paths once this flag is removed."
...
     ("math" . "clojure.math")
     ("set"  . "clojure.set")
     ("str"  . "clojure.string")
     ("walk" . "clojure.walk")
     ("zip"  . "clojure.zip"))
-  "Alist of aliases and namespaces used by `cljr-slash'."
+  "Alist of aliases to namespace libspec recommendations for `\\[cljr-slash]'.
+
+An optional keyword `:only` can limit a recommendation to the set of
+language contexts the libspec is available in."
```
vemv, line 104: Ideally, list all possible values of :only here

vemv (approved):
Nice ones, thanks!

clj-refactor.el
dgtized: That's up to the middleware, so once it's documented there we can link to accepted values. We can give clj/cljs as examples, but I don't think it makes sense to document the complete list from the client side.

vemv:
I've tagged 3.11.1

PR #559, opened 2025-01-18
use fallback for magic requires when there's no middlware

a13:
added a new defcustom `cljr-magic-requires-fallback` defcustom, so we can generate require expressions (from `cljr-magic-require-namespaces`) even if middleware is not available for some reason (e.g. when using babashka)
- also, I updated `cljr-magic-require-namespaces` with more [idiomatic aliases from The Clojure Style Guide](https://guide.clojure.style/#use-idiomatic-namespace-aliases).

### DIFF
--- a/clj-refactor.el
+++ b/clj-refactor.el
@@ -90,20 +90,45 @@ paths once this flag is removed."
   :type 'boolean
   :safe #'booleanp)
 
+(defcustom cljr-magic-requires-fallback t
+  "If t, `cljr-slash'' magic requires' will fallback to `cljr-magic-require-namespaces' if the middleware is not available."
+  :type 'boolean
+  :safe #'booleanp)
+
+;; see https://guide.clojure.style/#use-idiomatic-namespace-aliases
 (defcustom cljr-magic-require-namespaces
-  '(("edn"  . "clojure.edn")
-    ("io"     "clojure.java.io" :only ("clj"))
+  '(("async" . "clojure.core.async")
+    ("cli" . "clojure.tools.cli")
+    ("csv" . "clojure.data.csv")
+    ("datafy" . "clojure.datafy")
+    ("edn" . "clojure.edn")
+    ("io" "clojure.java.io" :only ("clj"))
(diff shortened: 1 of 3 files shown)


clj-refactor.el
```diff
@@ -90,20 +90,45 @@ paths once this flag is removed."
...
+    ("log" . "clojure.tools.logging")
+    ("mat" . "clojure.core.matrix")
     ("math" . "clojure.math")
-    ("set"  . "clojure.set")
-    ("str"  . "clojure.string")
+    ("p" . "clojure.core.protocols")
+    ("pp" . "clojure.pprint")
+    ("r" . "clojure.core.reducers")
+    ("s" . "clojure.spec.alpha")
+    ("set" . "clojure.set")
```
vemv, line 113: These almost certainly will conflict will real-world preferences/usaged from users out there, so I'd rather leave them untouched.

vemv (commented):
Hi @a13 !

Thanks for the PR.

From my side I literally don't have the time in my hands to do a careful review.

My first instinct would be not to like the change because there's a lot of smartness in the middleware, and I wouldn't want users to miss out on that.

It's also the overall philosophy of CIDER - require an active connection to use most of the features.

To make matters more complicated, babasha-nrepl doesn't use the traditional nrepl middleware design, so refactor-nrepl cannot work there.

I don't have a definitive opinion on how to proceed.

Perhaps one can simply observe that most time one has at least one repl open. And that one could write .cljc files that work with bb and JVM alike, with reader conditionals. So that would work and could be documented.

Cheers - V

bbatsov:
Thanks a lot for this, and sorry it sat for so long. We've since added the same graceful-degradation to `cljr-slash` (and `cljr-add-missing-libspec`) directly: when the middleware isn't available they now fall back to `cljr-magic-require-namespaces`, so common aliases still work offline. That landed in #585 and will be part of 4.0.

Since it's covered now I'm going to close this one, but I really appreciate you taking the time to put it together.

PR #583, opened 2026-07-10
Add cljr-require-alias-at-point

bbatsov:
`cljr-slash` only adds a require *as you type the `/`* after an alias, so it can't help code you pasted (where the `/` is already there). This adds a command that requires the unresolved alias of the symbol at point - point can be anywhere in the aliased symbol, e.g. on `str/join`.

It reuses `cljr-slash`'s resolution (so it works offline via `cljr-magic-require-namespaces`, and can add a missing library per `cljr-slash-add-missing-libs`), is bound to `ra`, and is listed in `clj-refactor-menu`.

Compile clean (`--warnings-as-errors`), lint clean, 76 buttercup specs pass (incl. an offline end-to-end test and the no-alias error case).

- [x] The commits are consistent with our contribution guidelines
- [x] The new code is not generating byte compile warnings (run `make compile`)
- [x] All tests are passing (run `make test`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Add `cljr-require-alias-at-point` (mnemonic `ra`): add a require for the unresolved namespace alias of the symbol at point, without retyping `/` - handy after pasting code that uses an alias which isn't required yet. It resolves the alias the same way `cljr-slash` does (works offline, and can add a missing library).
 - `cljr-slash` now works without a running REPL. When refactor-nrepl's `suggest-libspecs` op isn't available, it falls back to the static `cljr-magic-require-namespaces` table (honoring each entry's `:only` language context) instead of erroring, so common aliases like `str`/`io` still get required before you jack in.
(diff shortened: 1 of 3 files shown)


rrudakov:
@bbatsov isn't this new command the same as `cljr-add-missing-libspec`?

```text
cljr-add-missing-libspec is an autoloaded interactive
native-comp-function in ‘clj-refactor.el’.

It is bound to C-c RET a m.

(cljr-add-missing-libspec)

Inferred type: (function nil t)

Requires or imports the symbol at point.

If the symbol at point is of the form str/join then the ns
containing join will be aliased to str.

See: https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-add-missing-libspec
```

bbatsov:
Good catch - yeah, this overlaps with `cljr-add-missing-libspec`, which already handles `str/join` at point when connected. The only thing the new command adds is working without a REPL (via the static alias table), which is pretty niche. Going to revert it (#584) - thanks for flagging.

